PR #859, opened 2014-10-05
[KRAFDBCK-11099] Fix to AgendaCreatedNotificationRenderer

leo-at-rsmart:
#viewAgenda in MeetingActionsActionBase.java assumes that `scheduleIdFk` is the `ID` field in the `COMM_SCHEDULE` table; however, scheduleId refers to the `SCHEDULE_ID` field in the `COMM_SCHEDULE` table. These values may not (and for the most part aren't) always the same. This changes uses ID field which is intended.

fixes #KRAFDBCK-11099


jflach:
I cannot reproduce this issue in TRUNK-6ad58301 (Jan-13-02:18).

PR #1344, opened 2014-12-19
KRACOEUS-8453: Closing dialog window through document actions

jamey-rsmart:


jamey-rsmart:
@blackcathacker, discussed your suggestion for the simplified buttons with Victor and he likes it.  The only problem is that while the Rolodex maint doc that is currently used is the old KNS version, changing the KRAD footer here will change it globally. If the maint docs are ever to be switched over to KRAD, the Rolodex maint doc will have only two buttons.  This begs the question, is this even a concern anymore with the state of KRAD and KC moving forward?

blackcathacker:
First, no. Second, if we ever do move the Rolodex to using KRAD we can create a new KRAD maint doc view for the typical case and another for this one. So either way, you should be fine to make this custom maint doc work specifically for inline adding.

jamey-rsmart:
Made changes and rebased.

blackcathacker:
Looks much better. Thanks!

PR #1348, opened 2014-12-22
KRACOEUS-8526: refreshing pd user roles on navigate

leoherbie:


leoherbie:
@shojmr no because new roles may have been added giving different capabilities

leoherbie:
After thinking about this, the way I could optimize this is to go directly to the table backing the kim derived role to decide if things are dirty. I'd prefer not to do this unless we have determined a performance problem. I think in general the table backing the kim role should be internal.

PR #1355, opened 2014-12-23
KRACOEUS-8436 PD UXI: S2S Form Print - Select All selects all Available forms, not all Included forms

sasipolus:
Committed the files.


tdurkin:
What's the status on this?  Do we need to keep the JIRA open and to the client side refresh?  Or can we merge this and create a new JIRA to address the client side?

leoherbie:
I think we @blackcathacker and I would like this work done on the client side.

sasipolus:
Rebased and committed the files with client side change.

PR #1359, opened 2014-12-23
KULRICE-14083 isCurrentQuestionnaire refactor

cniesen:
The QuestionnaireServiceImpl.isCurrentQuestionnaire method does a table scan and loads all questionnaires of a specific questionnaireSeqId. 

In my local install the table has only 18 records and at most two revision.  The isCurrentQuestionnaire method gets called either 10 times (20 if Questionnaire page is reloaded, i.e. not navigated to another page of the document).

Initially I though a performance gain could be achieved but final before/after testing didn't show any significant change for my local installation.

I'm not sure how the data of the questionnaire table will grow in an actual KC implementation and therefore leave it up to you to make the final decision about applying this code change.

It's suggested to add the following index for best performance:
  alter table questionnaire add index (questionnaire_id, sequence_number desc);


blackcathacker:
test this please

cniesen:
I fixed the integration tests.  All should be good now.

PR #1361, opened 2014-12-26
KRACOEUS-8607

rmancher:
KRACOEUS-8607
Fix issue related to Employee and non-employee search
This fix works both in PD (multiple results returned based on wild card search) and returning multiple results in budget persons.

There is an issue with KRAD Light table where it binds only the first result based on search criteria. So only the first result will be returned if selected, otherwise no results are returned on a multi select search result based on search criteria. It works as expected in table collection.
Light table gives a better performance on searches with limited functionalities when compared to Table collection.

As per conversation with Brian, overriding Light table in KC until we get this fixed in rice.
Created a rice jira KULRICE-14122


blackcathacker:
So we copied all of light table from Rice just to modify a single method? Is there a reason we didn't just extend and override? Same with the light table DD override? Are all those property definitions new?

PR #1375, opened 2014-12-31
KRACOEUS-8621: adding missing rice location views

leoherbie:


blackcathacker:
So RIce doesn't provide these at all? Weird.

PR #1377, opened 2015-01-02
KRACOEUS-8129 fix fop error in: CommitteeBatchCorrespondenceServiceTest

boneypolus:
Committed the files.


kualiCoeusCi:
Can one of the admins verify this patch?

coeus-it/src/test/java/org/kuali/kra/test/infrastructure/lifecycle/KcIntegrationTestBaseLifecycle.java
```diff
@@ -36,7 +36,7 @@
     
     public static final String BROWSER_PROTOCOL = "http";
     public static final String BROWSER_ADDRESS = "127.0.0.1";
-    public static final String PORTAL_ADDRESS = "kc-dev";
+    public static final String PORTAL_ADDRESS = "";
```
blackcathacker: Why did you change this? That doesn't seem correct.

boneypolus: When PORTAL_ADDRESS = "kc-dev", we can't access the image using the relative path. Its always need kc-dev at first of the path. Since the template is using across the application we can't change the path to kc-dev. So we removed.

blackcathacker: And all other integration tests still pass?

blackcathacker:
test this please

coeus-it/src/test/java/org/kuali/kra/test/infrastructure/lifecycle/KcIntegrationTestBaseLifecycle.java
blackcathacker: @leoherbie can you look at this change? This is weird to me. Why do we even have this value if it isn't necessary?

leoherbie: I agree.  This doesn't seem right.  Why exactly don't the relative paths work?  I think this will break our other int tests

boneypolus: We tried other int tests after this change and its working fine.

leoherbie: did you run the full suite.  It doesn't take that long.  I'd suggest doing that.``` mvn clean install -Pitests```

rmancher:
@leoherbie thoughts on this?
@boneypolus remove PORTAL_ADDRESS if not required and run full suite to see if all tests are fine

blackcathacker:
I for one would still love to know why this change was required. Can you supply a diff of the changes you made to the stylesheet? And describe why the url matters to the fop and the stylesheet?

boneypolus:
Yes, we did changes in template. Corresponding table name is 'PROTO_CORRESP_TEMPL' , in this table we edited two rows of data to fix the error mentioned in jira.

I. Row with PROTO_CORRESP_TYPE_CODE = 20(DEFAULT-20-RenewalReminderLetter#1.xsl) , we did two changes,

1: To solve,
 ERROR org.apache.fop.apps.FOUserAgent - Image not found. URI: /export/home/www/https/tomcat5.0.25/webapps/coeus/images/couhes_byline2.gif. (No context info available) ,
we changed the url to static/images/couhes_byline2.gif and added a dummy image with the same name. Also removed PORTAL_ADDRESS ="kc-dev" from KcIntegrationTestBaseLifecycle.java

2: To solve, 
ERROR org.apache.fop.apps.FOUserAgent - Invalid property value encountered in column-width="proportional-column-width(9)": org.apache.fop.fo.expr.PropertyException: proportional-column-width() function may only be used when fo:table has table-layout="fixed".; property:'column-width' (No context info available) 
org.apache.fop.fo.expr.PropertyException: proportional-column-width() function may only be used when fo:table has table-layout="fixed".; property:'column-width'  ,
we added table-layout="fixed" in table definition.

II. Row with PROTO_CORRESP_TYPE_CODE = 23 (DEFAULT-23-ReminderToIrbNotification#1.xsl), we did one change to solve the same 1st fop error (image not found).

blackcathacker:
So this still makes me nervous. There must be something about the current directory when running jetty from these it-tests. There must be a better way to handle that than having the it-test jetty server startup with / being KC instead of using /kc-dev. I mean the change to remove kc-dev isn't serious, but at the same time its a hack to get this one test to work when we should understand why FOP is looking in the wrong place when running it-tests.

leoherbie:
I agree with @blackcathacker. Could you explain why this isn't an issue when the app is deployed under the context kc-dev which is essentially what is happening here.

rmancher:
@boneypolus, I see that you modified the template to use external image url as parameter to get the image path.
We are retaining portal address in this case - right?

boneypolus:
@rmancher , Yes we are retaining portal address as it is and tried all the scenarios and its working fine.

blackcathacker:
Much more understandable solution. Thank you.

leoherbie:
thanks for this.  it is a much better solution.

PR #1378, opened 2015-01-02
KRACOEUS-8555: Updated look and feel of medusa to be easier to read as well as between PropDev and the PropDev lookup.

jamey-rsmart:


coeus-webapp/src/main/webapp/css/common/core.css
```diff
@@ -298,4 +298,23 @@ li.kc-listIconDropdown li a.uif-boxLayoutHorizontalItem  {
...
+    color: #07c;
+    white-space:nowrap;
+    text-decoration:none;
+    padding:1px 2px;
+    margin:0;
+}
+div.jstree li a.IP:hover, div.jstree li a.DP:hover,
+div.jstree li a.irb:hover,div.jstree li a.iacuc:hover,
+div.jstree li a.neg:hover,div.jstree li a.award:hover,
+div.jstree li a.subaward:hover{
```
blackcathacker: Is there a reason we have all these classes? Can't we just use a global a.medusa or something? Also, the jstree usage here is a little concerning. I feel like we might use jstree outside of KRAD and outside of medusa. Is this styling intended to be global across the app? If not, maybe add a class to the jstree for medusa use that instead?

coeus-webapp/src/main/webapp/scripts/common/medusa.js
```diff
@@ -1,11 +1,36 @@
...
+    namespace.externalMedusaLoader = function() {
+        //minimizeMedusaTrees gets called before the jstree functionality.  set a timer to wait for it to be added.
+        var timer = undefined;
+
+        timer = setInterval(function() {
+            if ($('div.jstree li a')){
+                clearInterval(timer);
+                namespace.minimizeMedusaTrees();
+            }
+        },10);
```
blackcathacker: Yuck! But ok.

coeus-webapp/src/main/webapp/css/common/core.css
jamey-rsmart: good idea, didn't think about that.

jamey-rsmart: As far as the various classes, "DP, IP, irb, etc..." it's set up in the krad (treeNode.ftl) to add the nodeType as a class.  That's why there are all these classes, which in this implementation doesn't really mean much.

blackcathacker: Ahhh...ok. Does div.medusa li>a work instead of this ugly mess of css? If this is really the best way to do it, then I'm fine with it. Needs to be rebased regardless though.

blackcathacker:
Like this a lot more too. Thanks again.

PR #1387, opened 2015-01-05
KRACOEUS-8626: turning off client side validation for sync-all

leoherbie:


blackcathacker:
I think this problem is bigger than this. The sync and create hierarchy from children don't require the budget-type, but the link to parent requires both fields. I think the dialog should probably be redesigned...but its a start.

PR #1388, opened 2015-01-05
KRACOEUS-8632

rmancher:
KRACOEUS-8632
KRACOEUS-8631
Fix/Verify all budget print forms
Subaward NPE and cost element refresh issue - working after merging code this morning.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/common/budget/framework/personnel/BudgetPersonnelDetails.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/common/budget/framework/personnel/BudgetPersonnelDetails.java
@@ -177,6 +177,13 @@ public class BudgetPersonnelDetails extends BudgetLineItemBase implements Budget
     @Transient
     private Integer lineItemSequence;
 
+    /*
+     * Cost element BO is always referred from budget line item.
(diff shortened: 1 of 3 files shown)


coeus-impl/src/main/java/org/kuali/coeus/common/budget/framework/personnel/BudgetPersonnelDetails.java
```diff
@@ -636,7 +636,7 @@ public void setLineItemSequence(Integer lineItemSequence) {
 
     @Override
     public CostElement getCostElementBO() {
-        return costElementBO;
```
leoherbie: please remove the costElementBO since it is transient and seems to serve no useful purpose

rmancher: Yeah, I tried that since it doesn't make sense to have cost element in personnel details but it is defined in the base as abstract and used in various places. Removing that, I have to update code in few places. So didn't want to make last minute changes since I see that it is been used in calculation and other places.
Let me know.

rmancher:
@leoherbie, deprecated Cost element bo related.

PR #1393, opened 2015-01-05
KRACOEUS_8638: KRACOEUS-8638:narrative lock makes access readOnly, access page creat..

jwillia:
...s changes

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/auth/ProposalDevelopmentDocumentViewAuthorizer.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/auth/ProposalDevelopmentDocumentViewAuthorizer.java
@@ -1,11 +1,15 @@
 package org.kuali.coeus.propdev.impl.auth;
 
-import java.util.Set;
+import java.util.*;
 
 import org.apache.commons.lang3.StringUtils;
 import org.kuali.coeus.common.framework.auth.KcKradTransactionalDocumentViewAuthorizerBase;
 import org.kuali.coeus.propdev.impl.core.*;
 import org.kuali.coeus.propdev.impl.person.ProposalPerson;
+import org.kuali.coeus.sys.framework.service.KcServiceLocator;
+import org.kuali.coeus.sys.impl.lock.KcPessimisticLockService;
+import org.kuali.kra.authorization.KraAuthorizationConstants;
+import org.kuali.kra.infrastructure.Constants;
 import org.kuali.rice.kim.api.identity.Person;
 import org.kuali.rice.krad.document.Document;
 import org.kuali.rice.krad.document.authorization.PessimisticLock;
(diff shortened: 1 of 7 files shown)


blackcathacker:
The jira # is wrong in the commit.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentController.java
```diff
@@ -392,9 +393,12 @@ public ModelAndView getProposalAttachmentRights(@ModelAttribute("KualiForm") Pro
...
         List<NarrativeUserRights> editableRights = new ArrayList<NarrativeUserRights>();
         for (NarrativeUserRights right : selectedNarrative.getNarrativeUserRights()) {
-            NarrativeUserRights editableRight = new NarrativeUserRights();
-            PropertyUtils.copyProperties(editableRight,right);
-            editableRights.add(editableRight);
+            if (getKraAuthorizationService().hasPermission(right.getUserId(),form.getProposalDevelopmentDocument(), PermissionConstants.MODIFY_NARRATIVE) ||
+                    getKraAuthorizationService().hasPermission(right.getUserId(),form.getProposalDevelopmentDocument(), PermissionConstants.VIEW_NARRATIVE) ) {
+                NarrativeUserRights editableRight = new NarrativeUserRights();
+                PropertyUtils.copyProperties(editableRight,right);
+                editableRights.add(editableRight);
```
blackcathacker: I feel like this isn't completely correct. This will just correct the view of the permissions, but the narrative user rights value will still exist even though the user shouldn't have permission right? Seems like the user's NarrativeRights should be removed at the same time they are removed from having access to the proposal, instead of just hiding their permissions from view.

jwillia:
After further investigation there is already logic in place to remove narrative user rights when a row from access table is deleted.  However, since the access page is in the proposal development locking region the narratives (and the updated user rights) get over written by the narratives in the db, to be sure no concurrent narrative changes are lost.  

My proposed solution:

1.) Narrative Locks lock narrative and access page.  
2.) Access Page creates proposal development and narrative locks.
3.) When on access page the entire document gets saved. 

this should stop concurrent access modification if there are user editing access and narratives at the same time.

blackcathacker:
I think that reasonable. Is that fairly easy to accomplish?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/auth/ProposalDevelopmentDocumentViewAuthorizer.java
```diff
@@ -23,6 +27,8 @@
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 public class ProposalDevelopmentDocumentViewAuthorizer extends KcKradTransactionalDocumentViewAuthorizerBase {
 
+    private KcPessimisticLockService kcPessimisticLockService;
```
blackcathacker: Are these not Spring managed? I thought they were and could be autowired?

jwillia:
autowired services on the document authorizer

PR #1394, opened 2015-01-06
KRACOEUS-8633 : Allow subaward details to be viewed when readonly

blackcathacker:


coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/budget/subaward/ProposalBudgetSubawardsPage.xml
```diff
@@ -138,7 +141,7 @@
 					p:order="10" />
 				<bean parent="PropBudget-SubawardsAttachments-FileUpload-Section" 
 					p:hidden="@{dialogDataObject?.budgetSubAwardFiles == null ? false : !dialogDataObject.budgetSubAwardFiles.isEmpty()}" 
-					p:order="20" />
+					p:order="20" p:render="@{!#parent.readOnly}"/>
```
tharpgc: May be a stupid question, but why (in readonly mode) hide the file upload section here, but above you show a disabled save button? Is that a standard we are following?

coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/budget/subaward/ProposalBudgetSubawardsPage.xml
```diff
@@ -192,7 +195,7 @@
 					</property>
 				</bean>
 				<bean p:actionLabel="Sync from PDF" parent="Uif-SecondaryActionButton-Mini" p:order="30" 
-					p:disabled="@{dialogDataObject == null}" p:methodToCall="syncFromPdf">
+					p:disabled="@{dialogDataObject == null}" p:render="@{!#parent.readOnly}" p:methodToCall="syncFromPdf">
```
tharpgc: Ditto to above question. Why hide the button in readonly mode here, but show a disabled save button above?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/ProposalBudgetServiceImpl.java
```diff
@@ -265,11 +265,13 @@ public ProposalDevelopmentBudgetExt copyBudgetVersion(ProposalDevelopmentBudgetE
     
     protected ProposalDevelopmentBudgetExt copyBudgetVersionInternal(ProposalDevelopmentBudgetExt budget, DevelopmentProposal developmentProposal) {
         for (BudgetSubAwards subAwards : budget.getBudgetSubAwards()) {
```
leoherbie: could you please put a comment on this block of hacky get calls

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/hierarchy/ProposalHierarchyServiceImpl.java
```diff
@@ -854,9 +854,14 @@ protected void synchronizeChildBudget(DevelopmentProposal hierarchyProposal, Dev
             BudgetSubAwards newSubAwards;
             Map<Integer, BudgetSubAwards> subAwardMap = new HashMap<>();
             for (BudgetSubAwards childSubAwards : childBudget.getBudgetSubAwards()) {
+            	//pre-fetch all clobs
```
leoherbie: please expand this comment so we understand what is going on in the future

coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/budget/subaward/ProposalBudgetSubawardsPage.xml
blackcathacker: Hiding the button here and some related ones out of necessity as both criteria (dialogDataObject and readOnly) created JS errors as disabled is evaluated client side.

coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/budget/subaward/ProposalBudgetSubawardsPage.xml
blackcathacker: This section is hidden unless the user clicks a button(which will also be hidden), so not rendering it in this case seemed better. As far as a standard, no. But I prefer to show buttons as disabled instead of hiding them as it gives a better user experience. Instead of wondering how to do something you can't, you simply see that you can't do it.

blackcathacker:
Comments added/updated per review.

PR #1395, opened 2015-01-06
KRACOEUS-8634 : Fix removing budget line items for subaward

blackcathacker:


tharpgc:
+1

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/subaward/ProposalBudgetSubAwardController.java
```diff
@@ -199,14 +199,9 @@ public ModelAndView addSubAward(@ModelAttribute("KualiForm") ProposalBudgetForm
 	@Transactional @RequestMapping(params={"methodToCall=deleteLine", "actionParameters[selectedCollectionPath]=budget.budgetSubAwards"})
 	public ModelAndView deleteLine(@RequestParam("actionParameters[lineIndex]") Integer subAwardIndex, @ModelAttribute("KualiForm") ProposalBudgetForm form) {
 		BudgetSubAwards subAwardToDelete = form.getBudget().getBudgetSubAwards().get(subAwardIndex);
-		form.getBudget().getBudgetLineItems().removeAll(subAwardToDelete.getBudgetLineItems());
```
leoherbie: ha that doesn't work.  Line items are off the budgetPeriod and aren't mapped on the budget.  Related to this, should we be doing the "removeAll" from the periods rather than the deleteMatching you have below?  That would seem the more "proper" way

rmancher:
+1

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/subaward/ProposalBudgetSubAwardController.java
rmancher: @leoherbie, I see that deletematching is removed and currently it is removing line items from period. That is correct.
Are you referring to something different?

leoherbie: duh.  you are correct.  this looks fine

PR #1403, opened 2015-01-06
KRACOEUS-8646:fix test

jwillia:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/hierarchy/ProposalHierarchyServiceImpl.java
```diff
@@ -2043,6 +2044,9 @@ public void setKcDocumentRejectionService(
 	}
 
     public PessimisticLockService getPessimisticLockService() {
+        if (pessimisticLockService == null) {
+            pessimisticLockService = KcServiceLocator.getService(PessimisticLockService.class);
+        }
```
blackcathacker: Why are we doing this?

jwillia:
removed unneeded service locator call.

PR #1404, opened 2015-01-06
[KRACOEUS-8651] full name data attr for personnel

bsmith83:


blackcathacker:
I'm not a fan of commit messages with feedback queue jiras. Can we move these before committing?

blackcathacker:
@bsmith83 I moved the jira to kracoeus, can you amend the commit message please?

PR #1407, opened 2015-01-06
KRACOEUS-8654: Created error message to stop budget periods from being c...

jamey-rsmart:
...reated that have a start date before or equal to the last period's end date.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/period/BudgetPeriodRule.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/period/BudgetPeriodRule.java
@@ -340,6 +340,11 @@ public class BudgetPeriodRule {
                         newBudgetPeriod.setBudgetPeriod(periodNum+1);
                     }
                     break;
+                }else if(newPeriodStartDate.compareTo(periodEndDate) <= 0) {
+                    dateCompareValue = "ERROR_NEW_PERIOD_START_BEFORE_PREVIOUS_END";
+                    saveErrors(dateCompareValue, errorMap, newPeriodDateParams);
+                    validNewBudgetPeriod = false;
+                    break;
                 }
                 previousPeriodStartDate = budgetPeriod.getStartDate();
                 previousPeriodEndDate = budgetPeriod.getEndDate();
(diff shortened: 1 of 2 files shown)


tharpgc:
+1

coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/period/BudgetPeriodRule.java
```diff
@@ -340,6 +340,10 @@ private boolean isValidNewBudgetPeriod(Budget budget, BudgetPeriod newBudgetPeri
                         newBudgetPeriod.setBudgetPeriod(periodNum+1);
                     }
                     break;
+                }else if(newPeriodStartDate.compareTo(periodEndDate) <= 0) {
+                    dateCompareValue = "ERROR_NEW_PERIOD_START_BEFORE_PREVIOUS_END";
+                    saveErrors(dateCompareValue, errorMap, newPeriodDateParams);
+                    validNewBudgetPeriod = false;
```
blackcathacker: This doesn't seem correct to me. This is in a loop of all existing periods. If I've modified dates and I'm trying to insert a period between 2 existing periods, seems as though this will always throw an error incorrectly in that case.

jamey-rsmart: @blackcathacker, I've poured through the validation and the only thing that appears to be missing is a break to get out of the loop the moment this error is detected so other budget periods might not be falsely identified?  Is this a correct assumption?

blackcathacker: So if you edit the budget periods to create a gap between two existing budget periods. Say 1/1/2014-6/30/2014 and 1/1/2015-6/30/2015. And then try to insert a new budget period at 7/1/2014-12/31/2014. Does that work? Seems as though this change would cause that to fail when it shouldn't. Am I reading the logic here wrong?

jamey-rsmart: OK, I wrote this earlier but never actually submitted it.

This works last time I checked, though perhaps I didn't test this exact condition (pretty sure, but I'll do it again to be sure).  Since the periods are automatically sorted by the date, the new starting date 7/1/2014 would be compared to the first period ending at 6/30/2014 producing a 1 and fail the if condition.  Putting a break in will stop it from failing on the second period.  That I believe is needed (If nothing else than saving time from making unnecessary iterations through the loop).

blackcathacker: Ok....reading the code again I think you are correct. I was missing the part where the previous if would catch the second period and your case wouldn't. Although I also think you should put a break in there if you detect an error to avoid it looping further if there is a problem.

coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/period/BudgetPeriodRule.java
```diff
@@ -340,6 +340,11 @@ private boolean isValidNewBudgetPeriod(Budget budget, BudgetPeriod newBudgetPeri
...
+                    break;
                 }
                 previousPeriodStartDate = budgetPeriod.getStartDate();
                 previousPeriodEndDate = budgetPeriod.getEndDate();
```
jamey-rsmart: @blackcathacker break added.

PR #1410, opened 2015-01-07
KRACOEUS-8662:fixed delete permissions

jwillia:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentViewHelperServiceImpl.java
```diff
@@ -569,24 +560,6 @@ public String getQuestionMoreInfo(Question question) {
         return moreInfo.toString();
     }
 
-    @Override
-    public void retrieveEditModesAndActionFlags() {
-    	super.retrieveEditModesAndActionFlags();
-    	View view = ViewLifecycle.getView();
-        UifFormBase model = (UifFormBase) ViewLifecycle.getModel();
-    	if(! getKcWorkflowService().isInWorkflow(((DocumentFormBase) model).getDocument())) {
```
leoherbie: Could you explain why this needs to be removed in more detail?

jwillia: this is adding the delete permission to the action flags for all users regardless of permissions if the document is not in workflow.  The document authorizer is already checking that the document is not in work flow in addition to checking that the user has the delete proposal permission, so i second check here is ignoring the permissions in addtion to duplicating logic in the authorizer.

leoherbie: ok thanks for that

PR #1413, opened 2015-01-07
KRACOEUS-8630: Handle personnel removal while unlinking and syncing

gathreya:


jwillia:
:+1:

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/hierarchy/ProposalHierarchyServiceImpl.java
```diff
@@ -342,10 +344,44 @@ public void removeFromHierarchy(DevelopmentProposal childProposal) throws Propos
             }
             hierarchyProposal.setHierarchyOriginatingChildProposalNumber(lowestProposalNumber);
             dataObjectService.save(childProposal);
+            dataObjectService.save(hierarchyProposal);
```
blackcathacker: So I know I just merged this but, why are we saving here? Right after this we make all kinds of changes to the hierarchy proposal again? Should this happen after the sync or should we at least pass in the returns hierarchyProposal thats tied to the transaction?

gathreya: Sorry, just seeing this. So I actually thought some of the issues I was seeing was because the hierarchyProposal was not being saved in this sequence of code. I am working on more hierarchy stuff. I can try to take it out and test again.

PR #1414, opened 2015-01-07
KRACOEUS-8667:hide s2s elements once document is in workflow, and hide e...

jwillia:
...dit link on proposal search if proposal is in workflow

### DIFF
--- a/coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/core/DevelopmentProposal.xml
+++ b/coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/core/DevelopmentProposal.xml
@@ -1316,7 +1316,8 @@
     </bean>
 
     <bean id="Development-Proposal-Lookup-result-EditLink" parent="Uif-Link"
-          p:render="@{!#view.renderReturnLink}">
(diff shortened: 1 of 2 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/PropDevLookupableHelperServiceImpl.java
```diff
@@ -304,6 +306,11 @@ public void buildPropDevEditActionLink(Link actionLink, Object model,String titl
 
 	}
 
+    protected boolean isInWorkflow(String docId) throws WorkflowException{
+        String documentStatus = getWorkflowDocumentService().getDocumentStatus(docId).getCode();
```
leoherbie: Couldn't we pass in the PD doc which has the DocumentHeader where you can get the status.

jwillia: good idea

jwillia:
updated to hide edit link based on workflow document or the line.

PR #1419, opened 2015-01-07
KRACOUES-8637:only allow questionnaire copy if a questionnaire has answe...

jwillia:
...rs

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentViewHelperServiceImpl.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentViewHelperServiceImpl.java
@@ -30,6 +30,7 @@ import org.kuali.coeus.common.framework.custom.attr.CustomAttributeService;
 import org.kuali.coeus.common.framework.sponsor.Sponsor;
 import org.kuali.coeus.common.framework.sponsor.SponsorSearchResult;
 import org.kuali.coeus.common.framework.sponsor.SponsorSearchService;
+import org.kuali.coeus.common.questionnaire.framework.answer.Answer;
 import org.kuali.coeus.common.questionnaire.framework.answer.AnswerHeader;
 import org.kuali.coeus.common.questionnaire.framework.question.Question;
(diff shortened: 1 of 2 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentViewHelperServiceImpl.java
```diff
@@ -544,6 +545,19 @@ public boolean areActiveQuestionnaires(List<AnswerHeader> answerHeaders) {
         return false;
     }
 
+    public boolean areAnsweredQuestionnaires(List<AnswerHeader> answerHeaders) {
+        for (AnswerHeader answerHeader : answerHeaders) {
+            if (answerHeader.isActive()) {
+                for (Answer answer : answerHeader.getAnswers()) {
+                    if (StringUtils.isNotEmpty(answer.getAnswer())) {
```
leoherbie: I think this is wrong.  Don't all questions have to be answered (ie all answer objects have a non-blank answer).  Also is there existing code you can reuse that determines if a questionaire is complete?

jwillia: seeking more clarification in the jira

blackcathacker: Ohhhh...actually this looks reasonable. All this is doing is determining if the questionnaires can be copied to the new proposal. Not if the questionnaire is complete or not.

jwillia: per jira comment "If you can discern that any questions are answered but the questionnaire is not complete, then by all means, include the option to copy the questionnaire"

leoherbie: ok.  thanks for the clarification with functionals

PR #1420, opened 2015-01-07
KRACOEUS-8653

rmancher:
KRACOEUS-8653
Budget final version - updated to marked for submission.
Indicate whether budget version is marked for submission

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetViewHelperServiceImpl.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetViewHelperServiceImpl.java
@@ -23,6 +23,7 @@ import java.util.List;
 import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.kuali.coeus.common.api.sponsor.hierarchy.SponsorHierarchyService;
+import org.kuali.coeus.common.budget.framework.core.Budget;
 import org.kuali.coeus.common.budget.framework.core.BudgetAuditRuleEvent;
 import org.kuali.coeus.common.budget.framework.core.BudgetConstants;
 import org.kuali.coeus.common.budget.framework.distribution.BudgetCostShare;
@@ -30,6 +31,7 @@ import org.kuali.coeus.common.budget.framework.distribution.BudgetUnrecoveredFan
 import org.kuali.coeus.common.budget.framework.income.BudgetProjectIncome;
 import org.kuali.coeus.common.framework.ruleengine.KcBusinessRulesEngine;
 import org.kuali.coeus.common.impl.KcViewHelperServiceImpl;
+import org.kuali.coeus.propdev.impl.budget.ProposalBudgetService;
 import org.kuali.coeus.propdev.impl.budget.modular.BudgetModular;
 import org.kuali.coeus.propdev.impl.budget.modular.BudgetModularIdc;
 import org.kuali.coeus.propdev.impl.core.ProposalDevelopmentConstants;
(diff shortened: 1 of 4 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetViewHelperServiceImpl.java
```diff
@@ -205,4 +207,15 @@ public void setKcBusinessRulesEngine(KcBusinessRulesEngine kcBusinessRulesEngine
 		this.kcBusinessRulesEngine = kcBusinessRulesEngine;
 	}
     
+	public boolean isBudgetMarkedForSubmission() {
+		ProposalBudgetForm form = (ProposalBudgetForm)ViewLifecycle.getModel();
+		boolean budgetMarkedForSubmission = false;
+		Budget finalBudget = form.getDevelopmentProposal().getFinalBudget();
```
leoherbie: Seems like this could go in a service somewhere.  Then have the view helper service call that service.

rmancher: @leoherbie yeah thought of it, but then it is just one condition that we check. Not sure if there is any advantage in moving it to a service

leoherbie: At least pass in the Budget or DevelopmentProposal as an argument so we aren't relying on global memory which makes it easier to make out to a service at a later for other uses.

rmancher: Sure, have made those changes. Will check it in now

rmancher:
Addressed comments

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetViewHelperServiceImpl.java
```diff
@@ -205,4 +213,16 @@ public void setKcBusinessRulesEngine(KcBusinessRulesEngine kcBusinessRulesEngine
 		this.kcBusinessRulesEngine = kcBusinessRulesEngine;
 	}
     
+	public boolean isBudgetMarkedForSubmission() {
+		ProposalBudgetForm form = (ProposalBudgetForm)ViewLifecycle.getModel();
```
blackcathacker: I'm ok with this, but I dislike the useing ViewLifecycle. Global usages have plagued us. I'd prefer if you pass in the budget using SpEL @{budget}.

leoherbie: I'm in agreement with @blackcathacker and that was what I was trying to describe by may previous comment where I suggested passing the object in.

rmancher:
Removed ViewLifecycle
Code updated

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetViewHelperServiceImpl.java
```diff
@@ -205,4 +213,15 @@ public void setKcBusinessRulesEngine(KcBusinessRulesEngine kcBusinessRulesEngine
 		this.kcBusinessRulesEngine = kcBusinessRulesEngine;
 	}
     
+	public boolean isBudgetMarkedForSubmission(Budget finalBudget, Budget currentBudget) {
+		return getProposalBudgetService().isBudgetMarkedForSubmission(finalBudget, currentBudget);
```
blackcathacker: More than I was expecting to be passed in but looks good and SO happy to be rid of ViewLifecycle call.

PR #1422, opened 2015-01-08
KRACOEUS-8671: setting the budget status on the settings page now reflec...

leoherbie:
...ts on the versions page, canEditView is forced to reevaluate when setting the budget status

### DIFF
--- a/coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetSettings.xml
+++ b/coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetSettings.xml
@@ -30,10 +30,10 @@
 				<property name="items">
 					<list>
 						<bean parent="Uif-PrimaryActionButton" p:actionLabel="Apply Changes" p:methodToCall="saveBudgetSettings"
-							p:order="10" p:refreshId = "PropBudget-BudgetSettings-Dialog" p:dialogDismissOption=""
+							p:order="10" p:dialogDismissOption="REQUEST" p:performClientSideValidation="true"
 							p:render="@{!#parent.readOnly or #view.editModes.containsKey(T(org.kuali.coeus.propdev.impl.budget.core.ProposalBudgetConstants.AuthConstants).CHANGE_COMPLETE_STATUS)}"/>
 						<bean parent="Uif-SecondaryActionButton" p:actionLabel="Close"
-							p:methodToCall="closeBudgetValidation" p:order="20" />
+							p:methodToCall="closeBudgetValidation" p:order="20" p:performClientSideValidation="true" />
 					</list>
 				</property>
 			</bean>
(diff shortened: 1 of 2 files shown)


coeus-impl/src/main/resources/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetSettings.xml
```diff
@@ -30,7 +30,7 @@
 				<property name="items">
 					<list>
 						<bean parent="Uif-PrimaryActionButton" p:actionLabel="Apply Changes" p:methodToCall="saveBudgetSettings"
-							p:order="10" p:refreshId = "PropBudget-BudgetSettings-Dialog" p:dialogDismissOption=""
+							p:order="10" p:dialogDismissOption="REQUEST"
```
blackcathacker: I set the dialogDismissOption so that error messages will be displayed in the dialog. Doing this means those error either won't appear at all or will appear on the main page instead of in the settings.

leoherbie: The issue is that then the page doesn't refresh when the dialog is dismissed.  What do you suggest that we do?

leoherbie: or we could maybe use a different dialogDismissOption.  I'm not sure what all of them do but I think there is one that does all the validation first...

blackcathacker: I'm not sure. Maybe @bsmith83 can suggest something. Somehow we need to detect errors and not close the dialog in that case, if it succeeds we'll need to close the dialog and refresh the page. Maybe a callback when the dialog is destroyed that causes a refresh?

blackcathacker: Sadly no. I'm pretty sure all the dialogDismissOptions happen before the server round trip.

jwillia: can try something with Kc.Dialog.closeDialogWithoutError on success callback.  which checks for error css classes on the dialog before dismissing it.

leoherbie: I'm going to explore that.  thank @jwillia

rmancher: @leoherbie you can also try retrieveComponent to refresh in onHideDialogScript

bsmith83: @rmancher suggestion should work if I understand the question right

leoherbie:
After thinking about this, the way I could optimize this is to go directly to the table backing the kim derived role to decide if things are dirty.  I'd prefer not to do this unless we have determined a performance problem.  I think in general the table backing the kim role should be internal.

PR #1423, opened 2015-01-08
KRACOEUS-8506: Fixing JTA rollback exception

gathreya:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/person/KeyPersonnelServiceImpl.java
```diff
@@ -94,6 +94,15 @@ public  void setProposalPersonService (ProposalPersonService proposalPersonServi
      * @param document
      */
     public void populateDocument(ProposalDevelopmentDocument document) {
+        populateCreditSplit(document);
+
+        /* check for new certification questions */
+        for (ProposalPerson person : document.getDevelopmentProposal().getProposalPersons()) {
+            getYnqService().getPersonYNQ(person, document);
```
blackcathacker: So this was causing a rollback exception? Why?

gathreya: so in getPersonYNQ, we were changing YNQs which is apprently now causing a concurrentModification exception. In this navigation to the credit split tab situation, we do not really need to do this YNQ check anyway, just populate the credit split so I just do not do that anymore. Like I said, adding the @Transactional code in seems to have started this for some reason and I do not know why.

blackcathacker:
Looks like this is causing a compile exception in the unit tests.

gathreya:
oops, missed the second adapter.

gathreya:
test this please.

PR #1431, opened 2015-01-09
KRACOEUS-8655 : Check activity type change and empty rates

blackcathacker:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
```diff
@@ -68,7 +75,14 @@ public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam
         if (auditActivated != null){
             form.setAuditActivated(Boolean.parseBoolean(auditActivated));
         }
-        return getModelAndViewService().getModelAndViewWithInit(form, ProposalBudgetConstants.KradConstants.BUDGET_DEFAULT_VIEW);
+        ModelAndView result = getModelAndViewService().getModelAndViewWithInit(form, ProposalBudgetConstants.KradConstants.BUDGET_DEFAULT_VIEW);
```
leoherbie: put this line in the last else block

gathreya:
needs to be rebased now too

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
blackcathacker: I can't. This is required regardless because we need to do the 'withInit' part before doing anything else.

blackcathacker:
Rebased

coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/rate/BudgetRatesServiceImpl.java
```diff
@@ -173,6 +179,8 @@ protected void checkActivityPrefixForRateClassTypes(List<RateClassType> rateClas
         List<BudgetLaRate> budgetLaRates = budget.getBudgetLaRates();
         for(RateClassType rateClassType : rateClassTypes) {
             if(rateClassType.getPrefixActivityType()) {
+            	//making changes to the DO here, need to detach to make sure these changes aren't persisted.
+            	entityManager.detach(rateClassType);
                 String newRateClassTypeDescription = activityTypeDescription.concat(rateClassType.getDescription()); 
                 rateClassType.setDescription(newRateClassTypeDescription);
                 rateClassType.setPrefixActivityType(false);
```
gathreya: We do not want to persist this description change?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
leoherbie: oh god that is kind of terrible but oh well....

coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/rate/BudgetRatesServiceImpl.java
blackcathacker: No. Definitely not. The rateClassType is maint doc managed. It has a flag on it, prefixActivityType and if that is set then the description should be prefixed the with current budget activity type in the view, but if we persist that then the first budget created will modify that maint doc record and all future budgets will have that activity type for that rate class type.

Would be more correct to use a DTO here instead anyway, but this code is very complex and I didn't want to tweak it more than necessary.

blackcathacker:
Also comment added for the WithInit call.

coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/rate/BudgetRatesServiceImpl.java
gathreya: ok

PR #1433, opened 2015-01-09
KRACOEUS-8683:fixed STE when linking with no proposal selected

jwillia:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/hierarchy/ProposalHierarchyServiceImpl.java
```diff
@@ -284,6 +284,7 @@ public void linkToHierarchy(DevelopmentProposal hierarchyProposal, DevelopmentPr
         List<ProposalHierarchyErrorWarningDto> errors = new ArrayList<ProposalHierarchyErrorWarningDto>();
         if (hierarchyProposal == null) {
```
gathreya: wrong jira numer??

jwillia:
fixed jira #

PR #1434, opened 2015-01-09
KRACOEUS-8608 Validation error incorrectly shows when attachment type ex...

shojmr:
...ists

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentHelper.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentHelper.java
@@ -1,9 +1,15 @@
 package org.kuali.coeus.propdev.impl.attachment;
 
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections4.Predicate;
+import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.kuali.coeus.propdev.impl.abstrct.ProposalAbstract;
+import org.kuali.coeus.propdev.impl.core.ProposalDevelopmentDocument;
+import org.kuali.coeus.propdev.impl.core.ProposalDevelopmentDocumentForm;
 import org.kuali.coeus.propdev.impl.person.attachment.ProposalPersonBiography;
 import org.kuali.rice.krad.bo.Note;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
```diff
@@ -354,7 +355,7 @@ protected ModelAndView navigate(ProposalDevelopmentDocumentForm form, BindingRes
          boolean canEdit = form.isCanEditView();
          if (isNavigateAwayFromAttachment(navigateToPageId, form.getPageId())) {
              prepareLocks(form);
-             return narrativePageSave(form, canEdit);
+             return narrativePageSaveOnNavigate(form, canEdit);
```
blackcathacker: I don't think these 2 methods do anything significantly different. The jira specifically says that the problem occurs when you click save on the attachments page before navigating away. Are you sure the issue isn't because we only handle the different locking sections of the proposal on navigate and ignore it when we save?

gathreya:
needs to be rebased too

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
shojmr: I guess you meant methods narrativePageSave and narrativePageSaveOnNavigate, they are different in, how they handle changed objects. narrativePageSave reads the document and replaces attachments from form object that is good to handle the scenario in which user saves first and navigates away. But it fails to handle the case of user updates/add an attachment and navigates away without saving. narrativePageSaveOnNavigate method handles newly added or updated attachments and don't disturb the already persisted attachments. I think replacing the already persisted attachments (which include blob data) with form data - which may not have the blob data for all the attachments result in attachments getting wiped out.
I am not sure how the locking sections are handled.

shojmr:
rebased

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
blackcathacker: My concern is navigate and save should be exactly the same except the page that loads afterwards. So if this is a problem in one case it should be in both. I see the difference between the 2 methods, but in either case you are copying attachments from the form to a db loaded version, its just a question of which ones. So are ones that haven't been edited losing their attachment because of this copy or something else? And if so, why doesn't it occur when they click save?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
```diff
@@ -394,11 +395,21 @@ protected ModelAndView narrativePageSave(ProposalDevelopmentDocumentForm form, b
...
         if (canEdit) {
             if (new ProposalDevelopmentDocumentRule().processAttachmentRules(form.getProposalDevelopmentDocument())) {
-                //when saving on a page in the narrative locking region we don't want to over write proposal locking region data,
-                //so we retrieve the latest proposal from the db, and replace the current propopsal with that, and then copy the attachments
-                document.getDevelopmentProposal().setNarratives(form.getDevelopmentProposal().getNarratives());
-                document.getDevelopmentProposal().setInstituteAttachments(form.getDevelopmentProposal().getInstituteAttachments());
-                document.getDevelopmentProposal().setPropPersonBios(form.getDevelopmentProposal().getPropPersonBios());
+                for(Narrative narrative : form.getDevelopmentProposal().getNarratives() ) {
+                	if(narrative.getObjectId() == null || narrative.isUpdated()) {
+                		document.getDevelopmentProposal().getNarratives().add(narrative);
```
blackcathacker: Still a little confused about this. So what happens in the case where the narrative existed, but was updated? Won't it end up in this list twice? And that doesn't cause problems?

leoherbie:
@shojmr what is the status here?

shojmr:
@blackcathacker Please review.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentHelper.java
```diff
@@ -38,6 +44,163 @@ public void reset() {
...
+    protected List<Narrative> findNarrativesByType(List<Narrative> origNarratives, final String narrativeTypeCode) {
+    	final List<Narrative> narrativesByType = new ArrayList<Narrative>();
+    	CollectionUtils.forAllDo(origNarratives, new Closure<Narrative>() {
+
+			@Override
+			public void execute(Narrative input) {
+
+				if(input.getNarrativeTypeCode().equals(narrativeTypeCode)) {
+					narrativesByType.add(input);	
+				}
```
blackcathacker: Why Closure here to build a list, but CollectionUtils.find above? Additionally narrativeTypeCode is not unique as some narrativeTypes allow multiple attachments. Use moduleNumber instead which is guaranteed unique across a proposal.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentHelper.java
```diff
@@ -38,6 +44,163 @@ public void reset() {
...
+    public List<ProposalPersonBiography> findPersonBiosByObjectId(List<ProposalPersonBiography> origBios,final ProposalPersonBiography personBiography) {
+    	final List<ProposalPersonBiography> biosByPersonAndType = new ArrayList<ProposalPersonBiography>();
+    	CollectionUtils.forAllDo(origBios, new Closure<ProposalPersonBiography>() {
+
+			@Override
+			public void execute(ProposalPersonBiography input) {
+
+				if(input.getObjectId().equals(personBiography.getObjectId())) {
+					biosByPersonAndType.add(input);	
+				}
```
blackcathacker: Closure here too, when you use find up above. Lets use just one. And here you use objectId, which in general we don't make common use of in KC. Better I think to use the primary key which would be proposalPersonNumber and biographyNumber in this case.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentHelper.java
```diff
@@ -38,6 +44,163 @@ public void reset() {
...
+    	Narrative updatedNarrative = CollectionUtils.find(origNarratives, new Predicate<Narrative>() {
+
+			@Override
+			public boolean evaluate(Narrative input) {
+				if(input.getNarrativeTypeCode().equals(newNarrative.getNarrativeTypeCode())) {
+					return true;
+				}	
+				else {
+					return false;
+				}
```
blackcathacker: Using narrativeTypeCode here as well which isn't unique. See other comment.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentHelper.java
```diff
@@ -38,6 +44,163 @@ public void reset() {
...
+    	ProposalPersonBiography updatedBio = CollectionUtils.find(origBios, new Predicate<ProposalPersonBiography>() {
+
+			@Override
+			public boolean evaluate(ProposalPersonBiography input) {
+				if(input.getObjectId().equals(newBio.getObjectId())) {
+					return true;
+				}	
+				else {
+					return false;
+				}
```
blackcathacker: And objectId here. Prefer use of the primary key instead. And actually looking at this again, aren't these find methods essentially duplicates? In one case you are returning a unique result and in others a list, but in both cases you are basing the search on the same criteria.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentViewHelperServiceImpl.java
```diff
@@ -351,6 +351,10 @@ protected boolean performDeleteLineValidation(ViewModel model, String collection
                 getProposalDevelopmentPermissionsService().processDeletePermission(document, ((ProposalUserRoles) deleteLine));
             }
         }
+        
+        if(deleteLine instanceof ProposalDevelopmentAttachment) {
+        	((ProposalDevelopmentAttachment)deleteLine).setUpdated(true);
+        }
```
blackcathacker: If we are deleting the attachment do we need to worry about this? Isn't it just removed?

blackcathacker:
@jwillia You've spent a lot of time with the pessimistic locking sections, so can you take a minute to look this over and see if you see any problems.

jwillia:
I have been able to pull @shojmr code locally, and test that no data is lost in concurrent modification in different locking regions.  My understanding of how this works is that it removes any attachments from the document that are not on the form, then replaces any attachments from the document that may have been updated on the form. However, I don't clearly understand how this fixes the issue of the missing attachment data though.

shojmr:
Comments addressed.
@jwillia From what I understand, In case of save/navigate if we replace an already saved attachment object on document with attachment object from form the file data will be set to null (I think file stream is not maintained across requests) and eventually the persisted blob will be deleted.

leoherbie:
this thing is super hacky but we are at the point where we just need to pull the trigger and see if we can fix things.

tharpgc:
Yeah, I know... I didn't like passing the error handler in, but in the interest of time I put it in anyway.

PR #1438, opened 2015-01-12
KRACOEUS-7658:fixed routing to set the correct update user

jwillia:
not sure if i wired this in spring correctly, the autowiring wasn't working producing npe's and when i tried to set the WorkflowDoucment and GlobalVariable service in the Override spring beans that application wouldn't start,  using service locators in the the getters was the only method I could get to work.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/sys/framework/workflow/KcPostProcessor.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/sys/framework/workflow/KcPostProcessor.java
@@ -16,19 +16,55 @@
 package org.kuali.coeus.sys.framework.workflow;
 
 import org.kuali.coeus.sys.framework.service.KcServiceLocator;
-import org.kuali.rice.kew.framework.postprocessor.DocumentRouteStatusChange;
-import org.kuali.rice.kew.framework.postprocessor.ProcessDocReport;
+import org.kuali.rice.kew.api.action.ActionType;
+import org.kuali.rice.kew.framework.postprocessor.*;
 import org.kuali.rice.krad.service.PostProcessorService;
 import org.kuali.rice.krad.workflow.postprocessor.KualiPostProcessor;
+
+import java.util.List;
 
 /**
  * Extends the {@code KualiPostProcessor} to record the actual user performing an action on a workflow status change.
  */
-public class KcPostProcessor extends KualiPostProcessor {
+public class KcPostProcessor implements PostProcessor {
     
     @Override
     public ProcessDocReport doRouteStatusChange(DocumentRouteStatusChange statusChangeEvent) throws Exception {
         return ((PostProcessorService) KcServiceLocator.getService("kcPostProcessorService")).doRouteStatusChange(statusChangeEvent);
     }
 
+    @Override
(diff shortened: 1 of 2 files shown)


leoherbie:
I would like to talk about this PR before merging.  @jwillia please skype me.

leoherbie:
I'll summarize the analysis @jwillia and I did.  The issue is that we have customized the post processor service but that service is not being called.  This customization makes it so the system user (kr) is not set to the update user.  Previously @jwillia did a spring override to force rice to use our custom service.  This would make all documents that route have our custom behavior.  This works but isn't the ideal solution.

Rice already provides a mechanism to configure post process behavior.  We have a custom post processor in KC (KcPostProcessor).  For some unknown reason the ProposalDevelopmentDocument.xml was not configured to use the KcPostProcessor.  All that needs to be done for this PR is to make the PD doc use the KcPostProcessor and then reingest the ProposalDevelopmentDocument.xml file.  This should be a one line change.

jwillia:
switched the post processor for proposal development document to use the KcPostProcessor, had to add doRouteLevelChange and doActionTaken to KcPostProcessor to ensure the update_user is updated if doRouteStatusChange is never invoked during workflow.

tdurkin:
Sounds like the post processor is shared among our documents.  If that's the case, can you make sure that our other docs are still working with the changes that you're making?  Will there be an OJB/JPA conflict with that?

leoherbie:
@tdurkin This postprocessor is only configured on the following:

KcMaintenanceDocument
AwardDocument
ProposalDevelopmentDocument (with his change)

I don't think there would be any JPA/OJB issues and this PR's change has a pretty low impact because he is not configuring a parent doc (which I thought about suggesting BTW)

coeus-impl/src/main/java/org/kuali/coeus/sys/framework/workflow/KcPostProcessor.java
```diff
@@ -31,4 +33,15 @@ public ProcessDocReport doRouteStatusChange(DocumentRouteStatusChange statusChan
         return ((PostProcessorService) KcServiceLocator.getService("kcPostProcessorService")).doRouteStatusChange(statusChangeEvent);
     }
 
+    @Override
+    public ProcessDocReport doRouteLevelChange(final DocumentRouteLevelChange levelChangeEvent) throws Exception {
+        return ((PostProcessorService) KcServiceLocator.getService("kcPostProcessorService")).doRouteLevelChange(levelChangeEvent);
```
leoherbie: Interesting that these methods were never implemented before.... Seems like a gap

leoherbie: I think we should do the following:

1) make the KcPostProcessor implement the PostProcessor interface NOT extend the KualiPostProcessor
2) make all methods defer directly to the KcPostProcessorService

Without doing this some methods are using the KcPosProcessorService and some are using the KualiPostProcessorService.  This is confusing and not ideal from a spring customization standpoint.

jwillia:
updated with @leoherbie suggestions, tested on proposal, maintenance, and award documents didn't encounter any issues.

PR #1439, opened 2015-01-12
KRACOEUS-8669: Display parent's status in child

gathreya:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentViewHelperServiceImpl.java
```diff
@@ -725,6 +725,15 @@ protected boolean renderEditForPersonnelAttachment(String personId, DevelopmentP
         return true;
     }
 
+    public String getProposalParentStatus(DevelopmentProposal proposal) {
+        if (proposal.isChild()) {
+            ProposalDevelopmentDocument parentDoc = getProposalHierarchyService().getParentDocument(proposal.getProposalDocument());
```
leoherbie: This seems a bit wasteful to get the entire Document just to get the status' description.  I wonder what we can do.

I'd rather see:  

Maybe a custom JPQL DAO?

gathreya: Yes, I think I can add something to the DAO

gathreya:
test this please.

gathreya:
changed, just used an existing DAO method.

leoherbie:
after talking with @gathreya there will be other changes needed.

gathreya:
made the required changes

PR #1441, opened 2015-01-12
KRACOEUS-8686

rmancher:
Fix issue related to audit rules invoked and correspinding save in attachments
causing unique constraint violation.
Issue here is that on submit proposal, audit rules are invoked and if there are rules where we need to add additional attachments, we navigate to attachments page, add attachment and save is reinserting existing records causing unique constraint violation. 
While saving from narratives page, we retrieve the latest document from database and setting attachments to avoid concurrent modification issue. 
Not sure what is going on exactly here but looks like eclipse link is detaching the collection? and identifying it as new records and inserting

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
@@ -255,11 +255,12 @@ public abstract class ProposalDevelopmentControllerBase {
          ((ProposalDevelopmentViewHelperServiceImpl)form.getViewHelperService()).setOrdinalPosition(form.getDevelopmentProposal().getProposalPersons());
          saveAnswerHeaders(form, form.getPageId());
 
+         getTransactionalDocumentControllerService().save(form);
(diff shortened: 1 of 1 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
```diff
@@ -255,11 +255,12 @@ public ModelAndView save(ProposalDevelopmentDocumentForm form) throws Exception
          ((ProposalDevelopmentViewHelperServiceImpl)form.getViewHelperService()).setOrdinalPosition(form.getDevelopmentProposal().getProposalPersons());
          saveAnswerHeaders(form, form.getPageId());
 
+         getTransactionalDocumentControllerService().save(form);
```
jwillia: shouldn't we being running the audit rules before we save, or does it make a difference in this case?

rmancher: Yeah, it is causing problem when audit rules are executed before save on document that we retrieved at the time of narrative save.

PR #1442, opened 2015-01-12
KRACOEUS-8697: preventing a NPE

leoherbie:


leoherbie:
I need to look into why this issue is only happening in embedded but for now I want to fix the NPE to allow the creation of budgets.

PR #1443, opened 2015-01-12
[KRACOEUS-8416] Fix for fields that could potentially hidden by dialog h...

bsmith83:
...iding when validated on blur


coeus-webapp/src/main/webapp/scripts/common/global.js
```diff
@@ -262,4 +262,47 @@ KradResponse.prototype.updatePageHandler = function (content, dataAttr) {
...
\ No newline at end of file
+}
+
+/**
+ * Validate that a specific field's control defined by the selector/jQuery array passed in.  Also calls dependsOnCheck
+ * to validate any dependant fields.
+ *
+ * @param fieldControl selector/jQuery array that represents the control to validate
+ */
+function validateFieldValue(fieldControl) {
```
leoherbie: is this overriding something.  Is that why it's not in the Kc.Global object?

bsmith83: Yes

PR #1446, opened 2015-01-13
KRACOEUS-8688

rmancher:
KRACOEUS-8688
Fix optimistic lock exception and duplicate notes
We are not dealing with the latest notes which is causing optimistic locking.
Also this fix deals with duplicate entries in notes.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentDocument.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentDocument.java
@@ -611,8 +611,8 @@ public class ProposalDevelopmentDocument extends BudgetParentDocument<Developmen
 
     @Override
     public List getNotes() {
-        if (CollectionUtils.isEmpty(notes)) {
-            notes = new ArrayList<>(super.getNotes());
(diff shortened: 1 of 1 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentDocument.java
```diff
@@ -611,8 +611,8 @@ public NoteType getNoteType() {
 
     @Override
     public List getNotes() {
-        if (CollectionUtils.isEmpty(notes)) {
-            notes = new ArrayList<>(super.getNotes());
+        if (StringUtils.isNotBlank(getNoteTarget().getObjectId())) {
```
leoherbie: this seems like a lot of overhead.  Is there a way we can do this only if the CollectionUtils.isEmpty(notes) as well?

rmancher: Yeah, that is what I was thinking too and didn't want to put the fix right away even though I figured what is causing this issue.
I tried different scenarios and nothing worked.
Each time we add a note to the collection and save, we need to get the latest once we persist it.

PR #1451, opened 2015-01-13
KRACOEUS-8523 adjusting the last action request value so the ammendment...

jhulslander:
... routes

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/kra/irb/ProtocolDocument.java
+++ b/coeus-impl/src/main/java/org/kuali/kra/irb/ProtocolDocument.java
@@ -56,7 +56,10 @@ import org.kuali.kra.protocol.notification.ProtocolNotificationContextBase;
 import org.kuali.rice.coreservice.framework.parameter.ParameterConstants;
 import org.kuali.rice.coreservice.framework.parameter.ParameterConstants.COMPONENT;
 import org.kuali.rice.coreservice.framework.parameter.ParameterConstants.NAMESPACE;
+import org.kuali.rice.kew.actionrequest.ActionRequestValue;
+import org.kuali.rice.kew.actionrequest.service.ActionRequestService;
 import org.kuali.rice.kew.api.exception.WorkflowException;
+import org.kuali.rice.kew.service.KEWServiceLocator;
 import org.kuali.rice.krad.util.GlobalVariables;
 import org.kuali.rice.krad.util.ObjectUtils;
 import org.kuali.rice.krms.api.engine.Facts.Builder;
@@ -94,6 +97,8 @@ public class ProtocolDocument extends ProtocolDocumentBase {
     private static final String DISAPPROVED_CONTEXT_NAME = "Disapproved";
     
     private List<CustomAttributeDocValue> customDataList;
+    
+    private transient ActionRequestService actionRequestService;
     
     public ProtocolDocument() {
         super();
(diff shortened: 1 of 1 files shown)


coeus-impl/src/main/java/org/kuali/kra/irb/ProtocolDocument.java
```diff
@@ -163,6 +171,13 @@ protected void mergeAmendment(String protocolStatusCode, String type) {
...
         try {
             getDocumentService().saveDocument(newProtocolDocument);
+            //manually set the principleID of the last actionrequest to BLANK, so this routes
+            List requests = getActionRequestService().findAllActionRequestsByDocumentId(newProtocolDocument.getDocumentHeader().getDocumentNumber());
+            if (requests != null && requests.size() > 0) {
+            	ActionRequestValue lastRequest = (ActionRequestValue)requests.get(requests.size() -1);
+            	lastRequest.setPrincipalId("");
+            	getDataObjectService().save(lastRequest);
+            }
             // blanket approve to make the new protocol document 'final'
```
gathreya: Why are we doing this? So there are action requests which we are invalidating? What are these action requests?

jhulslander: The last action request, is the action the approved the merge of the protocol with the amendment.  If there is a problem with that, it would throw an exception before we get here.  By the time we get to this code, we just need to route the document to finalize it.

As per StandardWorkflowEngine.checkDefaultApprovalPolicy,  for the route to be approved, the document's initiator workflow ID, and the action request's principle ID must be different.  The code I have just resets the principle ID, so it will pass this test.

I think the ActionRequestValue in question here is being set by earlier processing in Rice.

coeus-impl/src/main/java/org/kuali/kra/irb/ProtocolDocument.java
```diff
@@ -163,6 +171,13 @@ protected void mergeAmendment(String protocolStatusCode, String type) {
         }
         try {
             getDocumentService().saveDocument(newProtocolDocument);
+            //manually set the principleID of the last actionrequest to BLANK, so this routes
+            List requests = getActionRequestService().findAllActionRequestsByDocumentId(newProtocolDocument.getDocumentHeader().getDocumentNumber());
+            if (requests != null && requests.size() > 0) {
+            	ActionRequestValue lastRequest = (ActionRequestValue)requests.get(requests.size() -1);
+            	lastRequest.setPrincipalId("");
+            	getDataObjectService().save(lastRequest);
```
leoherbie: This looks dubious but you should at least use:

```getActionRequestService().saveActionRequest(lastRequest)``` not ```getDataObjectService().save(lastRequest)```

File: coeus-impl/src/main/java/org/kuali/kra/irb/ProtocolDocument.java [hunk 3]
```diff
@@ -95,6 +99,10 @@
     
     private List<CustomAttributeDocValue> customDataList;
     
+    private transient DataObjectService dataObjectService;
```
leoherbie: you don't need DataObjectService

jhulslander:
I have removed the dataobjectservice call as @leoherbie requested.  Also concerning this being "dubious" ... I agree, but I couldn't find a better way of getting the PD amendment to route.  I am all hears for alternatives.

blackcathacker:
@jhulslander Can you explain why removing the person who took the last action fixes something? Seems as though the fix should be occurring in whatever code has a problem with a person taking that action, not in whether a person took it or not?

jhulslander:
The last action request, is the action the approved the merge of the protocol with the amendment. If there is a problem with that, it would throw an exception before we get here. By the time we get to this code, we just need to route the document to finalize it.

As per StandardWorkflowEngine.checkDefaultApprovalPolicy, for the route to be approved, the document's initiator workflow ID, and the action request's principle ID must be different. The code I have just resets the principle ID, so it will pass this test.

I think the ActionRequestValue in question here is being set by earlier processing in Rice.

blackcathacker:
So I would recommend changing the approval policy for the document then so we don't muck with the actionRequest.

leoherbie:
i agree with @blackcathacker

jhulslander:
Working on it .... thanks :-)

gathreya:
Closed since there is another PR to fix this.https://github.com/kuali/kc/pull/1515

PR #1454, opened 2015-01-13
[KRACOEUS-8695] Fix for proposal summary view

bsmith83:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDocumentSummaryView.java
```diff
@@ -0,0 +1,46 @@
...
+        setAdditionalScriptFiles(new ArrayList<String>());
+    }
+
+    protected DocumentEntry getDocumentEntryForView() {
+        DocumentEntry documentEntry = KRADServiceLocatorWeb.getDocumentDictionaryService().getDocumentEntryByClass(
+                ProposalDevelopmentDocument.class);
+
+        return documentEntry;
+    }
+}
```
gathreya: How did this break? Was this class not needed before?

bsmith83: This was broken by the AuthorizationCache stuff and some assumptions made there, so I had to add this workaround class to get around this.  The root problem and one we ran into when creating this view the first time is that you cant have two DocumentViews for the same document, this gets around that.

PR #1458, opened 2015-01-13
KRACOEUS-8470 managing validation boolean between award and budget, also...

jhulslander:
... fixing account type description validation on award


coeus-impl/src/main/resources/org/kuali/kra/datadictionary/Award.xml
```diff
@@ -1248,8 +1248,17 @@
     <bean id="Award-activityType.description" parent="ActivityType-description-parentBean"
           p:name="activityType.description" />
 
-    <bean id="Award-accountTypeDescription" parent="Award-accountTypeCode"
```
leoherbie: oops.  inheriting the wrong thing

PR #1477, opened 2015-01-15
KRACOEUS-8634: Handle subaward line items better

blackcathacker:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/subaward/BudgetSubAwards.java
```diff
@@ -114,9 +113,6 @@
     @OneToMany(mappedBy = "budgetSubAward", orphanRemoval = true, cascade = { CascadeType.ALL })
     @OrderBy("budgetPeriod")
     private List<BudgetSubAwardPeriodDetail> budgetSubAwardPeriodDetails;
-
-    @OneToMany(mappedBy = "budgetSubAward", cascade = {CascadeType.REFRESH, CascadeType.REMOVE})
-    private List<BudgetLineItem> budgetLineItems;
```
gathreya: huh, so you ended up removing this mapping and manage it manually?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/subaward/ProposalBudgetSubAwardController.java
```diff
@@ -199,8 +198,11 @@ public ModelAndView addSubAward(@ModelAttribute("KualiForm") ProposalBudgetForm
 	@Transactional @RequestMapping(params={"methodToCall=deleteLine", "actionParameters[selectedCollectionPath]=budget.budgetSubAwards"})
 	public ModelAndView deleteLine(@RequestParam("actionParameters[lineIndex]") Integer subAwardIndex, @ModelAttribute("KualiForm") ProposalBudgetForm form) {
 		BudgetSubAwards subAwardToDelete = form.getBudget().getBudgetSubAwards().get(subAwardIndex);
+		List<BudgetLineItem> lineItems = getDataObjectService().findMatching(BudgetLineItem.class, QueryByCriteria.Builder.fromPredicates(
+				PredicateFactory.equal("budgetId", subAwardToDelete.getBudgetId()), 
+				PredicateFactory.equal("subAwardNumber", subAwardToDelete.getSubAwardNumber()))).getResults();
```
rmancher: Why do we have to do this?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/subaward/BudgetSubAwards.java
leoherbie: It's really too bad that we can't have JPA handle this.  Is this because our object graph isn't organized well.  It seems that JPA should be able to keep track of these instances when they are removed from other places.  Do you think it is because krad-data is doing something?

blackcathacker: So previously we were mapping but not managing this relationship. So when new line items were added they weren't added here and attempting to add them here when creating them caused a problem because we don't want to persist this relationship since the BudgetPeriod technically owns the line items and JPA throws an exception in this case. And since SubAward doesn't actually own these child objects, I don't think orpahanRemoval is technically correct either so this mapping is unusual and complicated.

So we can keep this relationship and then have to call refreshReference on it whenever we need to use it which I hate and isn't really correct. We should strive to keep the object graph correct with the relationships. And I don't see how to do that in this case and I figured removing the relationship the most correct course.

PR #1479, opened 2015-01-15
KRACOEUS-8601:change hierarchy budget summary to only display latest bud...

jwillia:
...get


gathreya:
is this the wrong jira number?

jwillia:
yes it is

jwillia:
fixed the jira #

blackcathacker:
I don't think latest budget is correct here. The budget synced will either be the budget that synced manually by syncing from that budget, the budget for submission or the latest budget. Assuming the latest budget is likely the least correct of those 3. LastSyncedBudget is probably the one we should display in the hierarchy summary.

jwillia:
I have made the following changes.

1.) on hierarchy summary display the last synced budget if available, else display the final budget, else display latest budget.  All child proposals should have last synced budget but the hierarchy summary also displays parent proposals, which is why i included final and latest budgets in the logic.

2.) set the sync all button to refresh the summary page to automatically display the newly synced budgets.

PR #1480, opened 2015-01-15
KRACOEUS-8692: creating dialog for subawards warning, making null period...

leoherbie:
... display as ""


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetViewHelperServiceImpl.java
```diff
@@ -247,4 +248,24 @@ public ProposalBudgetService getProposalBudgetService() {
...
+    private boolean hasMultiplePeriods(Budget budget) {
+        return budget.getBudgetPeriods().size() > 1;
+    }
+
+    private boolean isBudgetLineItemExists(Budget budget) {
+        boolean lineItemExists = false;
+        for(BudgetPeriod budgetPeriod : budget.getBudgetPeriods()) {
+            if(budgetPeriod.getBudgetPeriod() > 1 && budgetPeriod.getBudgetLineItems().size() > 0) {
+                lineItemExists = true;
+                break;
```
blackcathacker: So if line items exist in any but the first period you don't need the warning? The warning seems to say it needs line items in all periods?

leoherbie:
The rule is if it has more than one period and is at least one of them empty then show the error.  I have fixed my sad logic.

PR #1486, opened 2015-01-16
KRACOEUS-8494 S2S Print forms: After first print attempt, no message given when printing all forms and there are errors on some forms that prevent them from being printed.

mrudulpolus:
Committed files.


leoherbie:
This PR depends on https://github.com/kuali/kc-s2sgen/pull/57 which has some api changes

mrudulpolus:
This PR has dependency with PR https://github.com/kuali/kc-s2sgen/pull/57

PR #1488, opened 2015-01-16
KRACOEUS-8726: switch to data object service

leoherbie:


leoherbie:
there is a fair amount of changes here but they are all pretty basic.  I figured we would give this a try and if there are any major issues we can roll it back

blackcathacker:
This is a LOT more than I was expecting. And I don't want to make this many changes that have little to no performance impact at what is hopefully only a couple of weeks before a release.

We can keep this PR/branch out there and apply these changes after 6.0 is out. Right now I'd rather see stability and 6.0 out the door than perfect performance.

In particular we aren't even sure if we'll be able to get EclipseLink L1 cache working due to the requirement that we use locking of some type to do that.

PR #1489, opened 2015-01-16
KRACOEUS-8713: Makes sure that a person has only one attachment per type

jamey-rsmart:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentControllerBase.java
```diff
@@ -394,11 +394,13 @@ protected ModelAndView proposalDevelopmentPageSave(ProposalDevelopmentDocumentFo
...
         ProposalDevelopmentDocument document = (ProposalDevelopmentDocument) getDocumentService().getByDocumentHeaderId(form.getDocument().getDocumentNumber());
         if (canEdit) {
-            if (new ProposalDevelopmentDocumentRule().processAttachmentRules(form.getProposalDevelopmentDocument())) {
+            if ((new ProposalDevelopmentDocumentRule().processAttachmentRules(form.getProposalDevelopmentDocument()))
+                    && (new ProposalDevelopmentDocumentRule().processPersonnelAttachmentDuplicates(form.getProposalDevelopmentDocument()))) {
             	form.getProposalDevelopmentAttachmentHelper().handleNarrativeUpdates(form, document);
             	form.getProposalDevelopmentAttachmentHelper().handleInstAttachmentUpdates(form, document);
             	form.getProposalDevelopmentAttachmentHelper().handlePersonBioUpdates(form, document);
                 document.getDevelopmentProposal().setProposalAbstracts(form.getDevelopmentProposal().getProposalAbstracts());
+                document.setDocumentNextvalues(form.getProposalDevelopmentDocument().getDocumentNextvalues());
```
blackcathacker: This is scary. First not related to this jira, secondly I'd also guess this means that this is an area of PD that can be modified in each locking section since I don't think attachments is the only thing that uses it. I think we need to do more here to merge the 2 lists when we do this.

jamey-rsmart: Which part is scary and which part is not related to the jira?

blackcathacker: So documentNextValues is used across prop dev. Person, person degrees, special review and attachments. By overwriting the value you are saying that the attachment section of the document owns documentNextvalues, when it doesn't. But the inverse isn't true as well, in that the other locking section doesn't own it either. So since both sections can modify this list, we need to do something smarter than just overwriting it. I think we need to merge the values basically.

jamey-rsmart:
Made changes.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/attachment/ProposalDevelopmentAttachmentHelper.java
```diff
@@ -69,7 +71,34 @@ public void handlePersonBioUpdates(ProposalDevelopmentDocumentForm form, Proposa
     	removeDeletedPersonBiosFromDocument(form.getDevelopmentProposal().getPropPersonBios(),document.getDevelopmentProposal().getPropPersonBios());
     	// Handle newly added and updated person bios
     	addNewUpdatePersonBiosOnDocument(form.getDevelopmentProposal().getPropPersonBios(),document.getDevelopmentProposal().getPropPersonBios());
-    	
+
+        //Extract the new document next values so they can be saved and not overwritten.
+        //document.setDocumentNextvalues(form.getProposalDevelopmentDocument().getDocumentNextvalues());
```
blackcathacker: Removed commented code.

jamey-rsmart: d'oh! ok, removing it now.

jamey-rsmart:
Changes made.

blackcathacker:
Hopefully this is a one-off, otherwise I'd rather see this in a separate service method. But since I think this is only place this will happen, looks good.

PR #1491, opened 2015-01-17
KRACOEUS-8693 : Fix budget person salary details mapping

blackcathacker:
Looking at it now more work might need to be done related to these details during hierarchy sync to account for budget period shift. But I'm not sure how critical that is since I can't find anywhere they are used in the actual calculation. But this fixes the duplication of these items in the child budget when the link was based on the id instead of the parent BO. This missing link would cause duplication of all items during sync into the child.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/personnel/BudgetPersonnelBudgetServiceImpl.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/personnel/BudgetPersonnelBudgetServiceImpl.java
@@ -118,6 +118,7 @@ public class BudgetPersonnelBudgetServiceImpl implements BudgetPersonnelBudgetSe
             for (BudgetPeriod budgetPeriodData : budgetPeriodList) {
                 BudgetPersonSalaryDetails personSalaryDetails = new BudgetPersonSalaryDetails();
               
+                personSalaryDetails.setBudgetPerson(budgetPerson);;
                 personSalaryDetails.setBudgetId(budget.getBudgetId());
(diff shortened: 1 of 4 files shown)


tdurkin:
I'm planning on merging this tonight unless someone else wants to comment first.  I'd rather have another PR next week to do more/clean-up if this will help avoid the performance issues we were having last week.

gathreya:
This pr looks fine. There is one empty statement in it that needs to be removed though.

coeus-impl/src/main/java/org/kuali/coeus/common/budget/impl/personnel/BudgetPersonnelBudgetServiceImpl.java
```diff
@@ -118,6 +118,7 @@ public void addBudgetPersonnelDetails(Budget budget, BudgetPeriod budgetPeriod,
             for (BudgetPeriod budgetPeriodData : budgetPeriodList) {
                 BudgetPersonSalaryDetails personSalaryDetails = new BudgetPersonSalaryDetails();
               
+                personSalaryDetails.setBudgetPerson(budgetPerson);;
```
tdurkin: @gathreya were you talking about this double ';'?

gathreya: Yes

gathreya: If @blackcathacker does not notice by end of day, let's merge nevertheless.

PR #1493, opened 2015-01-20
KRACOEUS-8727

rmancher:
KRACOEUS-8727
Fix months calculation in budget periods

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/ProposalBudgetNumberOfMonthsServiceImpl.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/ProposalBudgetNumberOfMonthsServiceImpl.java
@@ -16,118 +16,40 @@
 package org.kuali.coeus.propdev.impl.budget;
 
 
+import org.joda.time.DateTime;
+import org.joda.time.Months;
+import org.joda.time.Period;
+import org.joda.time.PeriodType;
+import org.kuali.coeus.sys.api.model.ScaleTwoDecimal;
 import org.springframework.stereotype.Component;
 
 import java.sql.Date;
-import java.util.Calendar;
-import java.util.GregorianCalendar;
 
 @Component("proposalBudgetNumberOfMonthsService")
 public class ProposalBudgetNumberOfMonthsServiceImpl implements ProposalBudgetNumberOfMonthsService {
-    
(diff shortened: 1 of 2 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/ProposalBudgetNumberOfMonthsServiceImpl.java
```diff
@@ -16,118 +16,40 @@
...
+
+    protected double getNumberOfMonths(Date startDate, Date endDate) {
+    	DateTime start = new DateTime(startDate);
+    	DateTime end = new DateTime(endDate);
+    	double daysInMonthForEndDate = end.dayOfMonth().getMaximumValue();
+    	
+    	PeriodType periodType = PeriodType.standard().withWeeksRemoved();
+    	Period period = new Period(start, end, periodType);
+    	double monthsBetween = Months.monthsBetween(start, end).getMonths();
+    	double daysLeft = period.getDays();
```
blackcathacker: I don't know joda time API well, but what happens if a period is longer than a year? I believe you can manually do that right?

blackcathacker:
Also that looks like a nice set of tests you are deleting. We can still use them or adapt them to work?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/ProposalBudgetNumberOfMonthsServiceImpl.java
rmancher: Yeah, we can edit and set the period to be longer than a year. 
Months are calculated accordingly.
Say start date : 01/01/2015 and end date 03/31/2016
Calculated months : 14.97

Result for linked JIRA
Say start date : 01/20/2015 and end date 01/13/2016
Calculated months : 11.77

rmancher:
@blackcathacker since we are not using any other methods that are in ProposalBudgetNumberOfMonthsServiceImpl service, I just removed it.
So I had to remove the same from the test as well. 
If it is just for the tests, I can leave it there in that service.

blackcathacker:
No. Removing code is good. And thanks for the answer to my joda question. Can you maybe add a unit test for the specific failing dates that were included in the jira since that seems so easy?

rmancher:
@blackcathacker just added couple of test with incorrect months based on start date
and end date mentioned in JIRA

PR #1496, opened 2015-01-20
KRACOEUS-8729-recall fix

gathreya:


gathreya:
test this please.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/core/ProposalDevelopmentDocument.java
```diff
@@ -249,13 +249,7 @@ public void doRouteStatusChange(DocumentRouteStatusChange dto) {
         if (!isProposalDeleted()) {
             DevelopmentProposal bp = this.getDevelopmentProposal();
             LOG.info(String.format("Route status change for document %s - proposal number %s is moving from %s to %s", bp.getProposalDocument().getDocumentHeader().getDocumentNumber(), bp.getProposalNumber(), oldStatus, newStatus));
-            if (bp.isParent()) {
-                try {
-                	getProposalHierarchyService().routeHierarchyChildren(this, dto);
-                } catch (ProposalHierarchyException e) {
-                    throw new RuntimeException("ProposalHierarchyException thrown while routing children.", e);
-                }
```
blackcathacker: This recall fix seems to remove significant hierarchy functionality? Does it not?

gathreya: Yes, we should not be routing child proposals anymore when the paren't is routed.

PR #1500, opened 2015-01-21
KRACOEUS-8728

rmancher:
KRACOEUS-8728
Fix to open budget in view mode when proposal is in view mode
Also in account return back to proposal.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
@@ -68,7 +68,8 @@ public class ProposalBudgetCommonController extends ProposalBudgetControllerBase
 
 	@MethodAccessible
 	@Transactional @RequestMapping(params="methodToCall=start")
-	public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam("auditActivated") String auditActivated, @ModelAttribute("KualiForm") ProposalBudgetForm form) {
+	public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam("viewOnly") String viewOnly, @RequestParam("auditActivated") String auditActivated, @ModelAttribute("KualiForm") ProposalBudgetForm form) {
+		boolean inViewMode = Boolean.parseBoolean(viewOnly);
 		form.setBudget(loadBudget(budgetId));
         getProposalBudgetLockService().establishBudgetLock(form.getBudget());
 		form.initialize();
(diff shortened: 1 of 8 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
```diff
@@ -68,7 +68,8 @@ public ModelAndView defaultMapping(@ModelAttribute("KualiForm") ProposalBudgetFo
 
 	@MethodAccessible
 	@Transactional @RequestMapping(params="methodToCall=start")
-	public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam("auditActivated") String auditActivated, @ModelAttribute("KualiForm") ProposalBudgetForm form) {
+	public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam("viewOnly") String viewOnly, @RequestParam("auditActivated") String auditActivated, @ModelAttribute("KualiForm") ProposalBudgetForm form) {
+		boolean inViewMode = Boolean.parseBoolean(viewOnly);
 		form.setBudget(loadBudget(budgetId));
         getProposalBudgetLockService().establishBudgetLock(form.getBudget());
```
blackcathacker: So view mode should affect the locking. Looks like we need more work on locking and readonly mode?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetForm.java
```diff
@@ -202,4 +203,12 @@ public void setDataValidationItems(List<DataValidationItem> dataValidationItems)
 		this.dataValidationItems = dataValidationItems;
 	}
 
+	public boolean isViewOnly() {
+		return viewOnly;
+	}
+
+	public void setViewOnly(boolean viewOnly) {
+		this.viewOnly = viewOnly;
+	}
```
blackcathacker: Do we really need an additional form property for this? Can't we do this using the view.readOnly or just form.canEditView?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
rmancher: In readonly/view can we skip locking?

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetForm.java
rmancher: Yeah, I tried that initially but I see that budget goes readOnly when it is set as complete.
So thought better not rely on that.

PR #1502, opened 2015-01-21
KRACOEUS-8593 updated the help files with  6.0 and 6.0 PD changes

jhulslander:


leoherbie:
@jhulslander will be taking a different approach where he deletes all the old content first before bringing in the new stuff.  This way, we are not leaving old, unused help in our codebase.

PR #1505, opened 2015-01-22
KRACOEUS-8723: fixing personnel attachments sync behavior

gathreya:


gathreya:
test this please.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/hierarchy/ProposalHierarchyServiceImpl.java
```diff
@@ -736,42 +740,48 @@ protected void syncAllPersonnelAttachments(DevelopmentProposal hierarchyProposal
...
-                destPropPersonBio.setVersionNumber(0L);
-                newList.add(destPropPersonBio);
 
+            // go over the child bios list and if person is not in multiple proposals add it
+            for (ProposalPersonBiography srcPropPersonBio : childProposal.getPropPersonBios()) {
+                // if the proposal is JUST being linked to a hierarchy and if this proposal has bios for
+                // people that exist on the parent, ignore those bios.
+                if (!isBioInNewChildDuplicate(isNewChild, hierarchyProposal, srcPropPersonBio)
+                    && personInMultipleProp.get(srcPropPersonBio.getPersonId()) != null &&
+                    !personInMultipleProp.get(srcPropPersonBio.getPersonId())) {
```
blackcathacker: Minor, but I never see anyone added to this list with a value of true. Could just check for existence or even better and just use a Set.

gathreya: Yeah, when I initially added it, I meant to use it for more than just a check but never ended up using it that way. I would rather not change this now though if it is minor. This code change itself I think risky at this point and if more hierarchy personnel attachments issues are filed, I am going to revert.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/hierarchy/ProposalHierarchyServiceImpl.java
```diff
@@ -331,7 +331,6 @@ public DevelopmentProposal removeFromHierarchy(DevelopmentProposal childProposal
             hierarchyProposal.setHierarchyOriginatingChildProposalNumber(lowestProposalNumber);
             dataObjectService.save(childProposal);
             dataObjectService.save(hierarchyProposal);
-            synchronizeAllChildren(hierarchyProposal);
             return childProposal;
```
gathreya: One thing I want to point on in this PR @blackcathacker is I removed syncAll when unlinking from hierarchy since I do not think it is needed at this stage and could make this slow when all someone is trying to do is remove a proposal from hierarchy. If you think we need to keep this in for some reason, let me know and I will add it in and test it again. Does budget unlinking depend on this somehow?

blackcathacker: No. I think thats ok. I think thats the way the code was originally written, but you are correct that it shouldn't be necessary to do this.

PR #1507, opened 2015-01-22
KRACOEUS-8678: Missing warning about invalid file type - Use different i...

tharpgc:
...nterface -- needed to switch to FileMeta interface class in method parameters, since I needed to check against file content type instead of file type.


gathreya:
was the jira reopened?

jflach:
It hadn't been resolved yet

PR #1514, opened 2015-01-23
KRACOEUS-8728

rmancher:
KRACOEUS-8728
Fix to NOT popup budget sync rates dialog when in view mode
Do not display lock message in view mode - both PD and Budget
Also as per PR comments earlier, I have set to establish locking when NOT in view mode. 
I did not see any issue at this point but thought we don't have to when budget is in view only mode.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
@@ -71,7 +71,10 @@ public class ProposalBudgetCommonController extends ProposalBudgetControllerBase
 	public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam("viewOnly") String viewOnly, @RequestParam("auditActivated") String auditActivated, @ModelAttribute("KualiForm") ProposalBudgetForm form) {
 		boolean inViewMode = Boolean.parseBoolean(viewOnly);
 		form.setBudget(loadBudget(budgetId));
-        getProposalBudgetLockService().establishBudgetLock(form.getBudget());
+		if(!inViewMode) {
+	        getProposalBudgetLockService().establishBudgetLock(form.getBudget());
+		}
+
 		form.initialize();
         if (auditActivated != null){
             form.setAuditActivated(Boolean.parseBoolean(auditActivated));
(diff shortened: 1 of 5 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
```diff
@@ -80,9 +83,9 @@ public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam
         ModelAndView result = getModelAndViewService().getModelAndViewWithInit(form, ProposalBudgetConstants.KradConstants.BUDGET_DEFAULT_VIEW);
         form.setViewOnly(inViewMode);
         form.setCanEditView(!inViewMode);
-        if (budgetRatesService.checkActivityTypeChange(form.getBudget().getBudgetRates(), form.getDevelopmentProposal().getActivityTypeCode())) {
+        if (!inViewMode && budgetRatesService.checkActivityTypeChange(form.getBudget().getBudgetRates(), form.getDevelopmentProposal().getActivityTypeCode())) {
```
blackcathacker: So I wonder if this viewMode idea needs to be more global. What happens if they only have view permission? Or if the budget is complete and therefore read-only?

rmancher: @blackcathacker  good catch. 
Initially I tried to set it up with view.readOnly which should work for all possible options.
I saw some issues/conflict with budget settings and to set budget in view mode when PD is opened in view mode.
Looking into it to see if that will still work as a global solution.

rmancher:
@blackcathacker I have made some changes to invoke confirmation dialog based on authorization.

coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetCommonController.java
```diff
@@ -80,9 +83,14 @@ public ModelAndView start(@RequestParam("budgetId") Long budgetId, @RequestParam
         ModelAndView result = getModelAndViewService().getModelAndViewWithInit(form, ProposalBudgetConstants.KradConstants.BUDGET_DEFAULT_VIEW);
         form.setViewOnly(inViewMode);
         form.setCanEditView(!inViewMode);
-        if (budgetRatesService.checkActivityTypeChange(form.getBudget().getBudgetRates(), form.getDevelopmentProposal().getActivityTypeCode())) {
+    	boolean canModify = false;
+    	if(!inViewMode) {
+    		canModify = getProposalBudgetAuthorizer().canModifyBudget(form.getBudget(), getGlobalVariableService().getUserSession().getPerson());
```
blackcathacker: So KIM checks are already run, probably multiple times and can be expensive. You should be able to just check for this in the editModes on the form?

rmancher: I don't think KIM checks are run at this point. Edit modes are not available.

blackcathacker: Hmm...ok. Thats frustrating but ok at least its only when we open the doc.

PR #1515, opened 2015-01-26
KRACOEUS-8523

rmancher:
KRACOEUS-8523
Blanket approve merged protocol document - IRB and IACUC.
Issue in this JIRA is that the new document not approved.
Status was ENROUTE not FINAL
Fixed in both IRB and IACUC

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/kra/protocol/ProtocolDocumentBase.java
+++ b/coeus-impl/src/main/java/org/kuali/kra/protocol/ProtocolDocumentBase.java
@@ -168,8 +168,7 @@ public abstract class ProtocolDocumentBase extends KcTransactionalDocumentBase i
             // after merge is done, then reset to the original usersession person.  
             // this is workaround for async.  There will be rice enhancement to resolve this issue.
             try {
-                DocumentRouteHeaderValue document = KcServiceLocator.getService(RouteHeaderService.class).getRouteHeader(
-                        this.getDocumentHeader().getWorkflowDocument().getDocumentId());
+                DocumentRouteHeaderValue document = getDocumentRouteHeaderValue();
                 String principalId = document.getActionsTaken().get(document.getActionsTaken().size() - 1).getPrincipalId();
                 String asyncPrincipalId = GlobalVariables.getUserSession().getPrincipalId();
                 String asyncPrincipalName = GlobalVariables.getUserSession().getPrincipalName();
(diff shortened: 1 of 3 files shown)


jhulslander:
+1

coeus-impl/src/main/java/org/kuali/kra/protocol/ProtocolDocumentBase.java
```diff
@@ -206,7 +205,16 @@ else if (isDisapproved(statusChangeEvent)) {
...
     
+    protected DocumentRouteHeaderValue getDocumentRouteHeaderValue() {
+        return KcServiceLocator.getService(RouteHeaderService.class).getRouteHeader(
+                this.getDocumentHeader().getWorkflowDocument().getDocumentId());
+    }
   
+    protected void setProtocolDocumentToApproveByDefault() {
+        DocumentRouteHeaderValue document = getDocumentRouteHeaderValue();
+        document.getDocumentType().getDefaultApprovePolicy().setPolicyValue(true);
+    }
```
gathreya: @rmancher this has always been false for protocol right? Why is this change needed now?

leoherbie: Can't we do this in the workflow xml?

rmancher: It is only for the merged Protocol document (amendment/renewal) where we need to blanket approve the new protocol document.

rmancher: @leoherbie as mentioned earlier, it is only for this specific scenario where we merge the amendment/renewal details to the new document once amendment/renewal is approved and we blanket approve the new document.

gathreya: @rmancher OK but was this not working in 5.2.1 then?

jhulslander: this does work in 5.2 (dly)

rmancher: @leoherbie yes it is working in 5.2.1
I'm not sure if anything changed with the default blanket approval policy with new rice upgrade.

gathreya:
OK, I do see the rice change that has necessitated the change of the approval policy. I am OK with these changes.

gathreya:
will merge if I do not hear from anyone else.

PR #1518, opened 2015-01-26
KRACOEUS-8210 institute rate and institute LA fields editable and audit ...

jhulslander:
...rules now prevent duplicates

in the DD and repository.xml I switched objectId to be the PK and removed the composite fields as the PK so the fields could be editable.  However, the database still has the composite fields as the PK.  I added objectId to the inquiry as it needs a unique field to work.  I took this path to avoid adding an extra field.  The approach also doesn't seem to create the problems Chuck had in budget finding the updates.  (Sync rates seems to work fine after an update happens)


tharpgc:
Well, that certainly seems to be simpler than my approach.  What does everyone else think about using the objectid as the PK?  I did not think that this was acceptable.  Also, did you try adding a new Institute Rate?  That was not causing me problems but we want to make sure your approach works with additions too.

jhulslander:
I created a PD, and a budget, went to the rates tab, saw it populated (to the best of my knowledge) correctly, and I changed a rate locally to the PD.
I then went to Institute Rate, and changed the rate I did a local PD edit for above.
I then opened up the PD budget, saw the previous rates were all there, then I synched rates, and saw my changes were pulled in.

Concerning the objectid as PK, that is hidden from the user, so from their perspective it is the same.

leoherbie:
We don't want to merge this work so close to release.  Thanks for all the hard work @jhulslander

jhulslander:
I grabbed the patch associated with this work and attached it to the Jira if this needs to be revisited in the future.

tdurkin:
Thanks, Jay!

PR #1519, opened 2015-01-26
KRACOEUS-8749: release scripts

gathreya:


gathreya:
@aakers please verify

PR #1528, opened 2015-01-28
[KRACOEUS-8622] remove location overrides from KC

bsmith83:


gathreya:
You've made sure this works after the rice upgrade right?

gathreya:
@tdurkin should we take this in?

tdurkin:
Do we still have to fake-out Remote mode be setting it to Local for embedded to work?

gathreya:
we should not have to after the rice upgrade. I have not checked. @bsmith83 can you?

leoherbie:
we should merge this if we can

bsmith83:
As far as I can tell it works but Im not sure where you guys were using these views, going directly to the lookup manually works for looking up

bsmith83:
I dont know what this means "Do we still have to fake-out Remote mode be setting it to Local for embedded to work?"

bsmith83:
I also noticed a small typo that persisted from the copy paste from the original files into rice, postal code lookup says county lookup in the header.  This is how you guys had it before though.

leoherbie:
I dont know what this means "Do we still have to fake-out Remote mode be setting it to Local for embedded to work?"

No we should not have to fake things out.  We do need to confirm this in our environment config.  I'm going to merge this for now and we can verify things in the morning.

PR #1532, opened 2015-01-29
KRACOEUS-8664:removing ignore proposalcopyservicetest, and monitoring ci

jwillia:


jwillia:
can not reproduce test error locally, testing if error still occurs in ci environment

PR #1531, opened 2015-01-29
KRACOEUS-8760:add application footer with copy right info

jwillia:

PR #1537, opened 2015-01-30
KRACOEUS-8760:add application footer with copy right info

leoherbie:


leoherbie:
this PR is a rebased version of https://github.com/kuali/kc/pull/1531

PR #1543, opened 2015-01-30
KRACOEUS-8760:moved version and sql info to footer

jwillia:
per @blackcathacker request moved version and sql version to footer


jwillia:
footerversion

blackcathacker:
Much better than in the header. Not sure about the positioning though. Maybe smaller text below and centered? I think we are finished with 6.0 in any case so this will probably be first thing in 6.0.1.

gathreya:
test this please.

jwillia:
footer2
updated to place on bottom

leoherbie:
why does it say NO VERSION?

jwillia:
thats what the version is set to in the config
<param name="build.version" override="false">NO VERSION</param>

leoherbie:
I see.  ok

PR #1555, opened 2015-02-03
KRACOEUS-8777:added hierarchy summary to budget module

jwillia:


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/budget/core/ProposalBudgetViewHelperServiceImpl.java
```diff
@@ -269,4 +275,26 @@ private boolean isEmptyBudgetLineItemExists(Budget budget) {
...
+
+    public void prepareHierarchySummary(ProposalBudgetForm form) {
+        if (form.getDevelopmentProposal().isInHierarchy()) {
+            form.setHierarchyDevelopmentProposals(getProposalHierarchyService().getHierarchyProposals(form.getDevelopmentProposal()));
+
+            for (DevelopmentProposal developmentProposal : form.getHierarchyDevelopmentProposals()) {
+                for (ProposalDevelopmentBudgetExt budget : developmentProposal.getBudgets()){
+                    if (budget.getBudgetSummaryDetails().isEmpty()){
+                        getBudgetCalculationService().populateBudgetSummaryTotals(budget);
+                    }
```
blackcathacker: Do we need to summarize all the budgets here? Can't we just do the last synced/final/newest?

jwillia: you are correct, this is a reminant from when we displayed all budgets on the summary, i will update it

jwillia:
updated to only calculate budget summary for budget used in the hierarchy summary, also made some updates so the budget summary is updated and refreshed if calling hierarchy actions from the submit page

PR #1557, opened 2015-02-03
KRACOEUS-8782:updated access collection to add with dialog, fixed check ...

jwillia:
...that roles have changed


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/auth/perm/ProposalDevelopmentPermissionsService.java
```diff
@@ -44,4 +44,6 @@
     public void processDeletePermission(ProposalDevelopmentDocument document, ProposalUserRoles proposalUser);
 
     public void processUpdatePermission(ProposalDevelopmentDocument document, ProposalUserRoles proposalUser);
+
+    public String getFullName(String first, String middle, String last);
```
leoherbie: this doesn't seem like the best place for this as it doesn't have much to do with permission other than it is used on the permissions page

jwillia: the method is used in the getPermissions method in the ProposalDevelopmentPermissionsService.  Which retrieve the workingUserRoles for the form.  I needed to expose the method in the service so that I can add the fullname to the collection for new line items, so that the dirty comparison would be comparing the same strings.  I didn't want to duplicate the logic or have the permissions service rely on a view helper service.

PR #1558, opened 2015-02-03
KRACOEUS-8776: lic header issue

leoherbie:


coeus-webapp/src/main/webapp/WEB-INF/tags/award/awardFandARates.tag
```diff
@@ -15,7 +15,7 @@
    - 
    - You should have received a copy of the GNU Affero General Public License
    - along with this program.  If not, see <http://www.gnu.org/licenses/>.
---%
+--%>
```
tdurkin: Really weird that this would only occur in one place, no?  Any way to make sure that we haven't missed any other places where this is screwed up?

leoherbie: I'll take a look

leoherbie: Doing the following regex search on our codebase only finds this single issue:

```--%[^>]```

tdurkin:
wo weird - but thanks for checking :-)

PR #1560, opened 2015-02-04
KRACOEUS-8783:allow routing actions for proposals accessed through propo...

jwillia:
...sal search

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/basic/ProposalDevelopmentHomeController.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/basic/ProposalDevelopmentHomeController.java
@@ -293,7 +293,8 @@ public class ProposalDevelopmentHomeController extends ProposalDevelopmentContro
                propDevForm.setDefaultOpenTab(defaultOpenTab);
            }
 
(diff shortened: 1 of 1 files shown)


coeus-impl/src/main/java/org/kuali/coeus/propdev/impl/basic/ProposalDevelopmentHomeController.java
```diff
@@ -293,7 +293,8 @@ public ModelAndView docHandler(@ModelAttribute("KualiForm") DocumentFormBase for
                propDevForm.setDefaultOpenTab(defaultOpenTab);
            }
 
-            if (StringUtils.equals(form.getRequest().getParameter("viewDocument"),"true") || StringUtils.equals(viewOnly, "true")) {
+            if (!propDevForm.getDocument().getDocumentHeader().getWorkflowDocument().isEnroute()
+                    && (StringUtils.equals(form.getRequest().getParameter("viewDocument"),"true") || StringUtils.equals(viewOnly, "true")) ) {
```
blackcathacker: This is weird. I think I can see why we need to do this, but I'd love it if we can not do this in the future. :)

PR #1570, opened 2015-03-16
release

leoherbie:


pom.xml
```diff
@@ -2012,12 +2012,12 @@
     <profiles>
         <profile>
             <id>grm</id>
-            <!-- comment out activation section to turn this profile off permanently-->
+            <!-- comment out activation section to turn this profile off permanently
```
blackcathacker: Ugh! We have to do this every time we "release"?

leoherbie: either that or tell them to build with mvn clean install -Dgrm.off each and every time

jhulslander: it would be cool to not have to do that

PR #1575, opened 2015-04-08
Uh fix pull request branch

gouldner:
Fixed several issues with Data Conversion RoleMember -> document_access conversion

I emailed Travis and Geo about these changes.  I also mentioned to Terry and he asked me to submit a pull request.

1) after converting only the KRIM_ROLE_MBR_ATTR_DATA_T entries were removed leaving KRIM_ROLE_MBR_T entries with no attribute data.  I altered to remove the role member from both tables.
2) the compare method used to filter duplicate document access role members was incorrect causing data loss.  The compare was only comparing document number so for a given role if more than one principal was a member only one would get converted and the others would be lost.  I corrected the compare to fix this issue while still removing true duplicates.
3) We found role member records in our database with null document id attribute.  So I added a check to ignore these.

NOTE: I accidentally commited a copy of the modified file in the second commit but removed it in the third commit.


leoherbie:
The contribution to the document_access conversion has been accepted.  It will be in our next release.  The contribution to the prop dev lookup are being actively worked on as well.  Thank you for the contribs.  Sorry it took so long to integrate them.

gouldner:
No problem glad to help.

PR #1576, opened 2015-04-16
[KRACOEUS-7230] Re-implementing 5.2.1 QuestionResolver changes in KC 6.x

rojlarge:
KRACOEUS-7230 made some changes to the QuestionResolver so that it would match Question Term Parameters to Questions based on Question ID and Questionnaire ID instead of Question ID and Questionnaire Ref ID. Those changes were included in KC 5.2.1, but it looks like they didn't make it into 6.0. This commit reimplements those changes, as well as a few other improvements/bug fixes (use sequence number when retrieving answer headers, only use latest version of questionnaires for answer resolution).

This code also requires the following SQL to update Question Term Param names to the new "Seq ID" format:
```
update KRMS_TERM_RSLVR_PARM_SPEC_T set nm = 'Question Seq ID' where nm = 'Question ID';
update KRMS_TERM_RSLVR_PARM_SPEC_T set nm = 'Questionnaire Seq ID' where nm = 'Questionnaire ID';
update KRMS_TERM_PARM_T set nm = 'Question Seq ID' where nm = 'Question ID';
update KRMS_TERM_PARM_T set nm = 'Questionnaire Seq ID' where nm = 'Questionnaire ID';
```

I wasn't sure where the best place for this was, since all the upgrade scripts except for 6.0.1 seem to have been removed from the Kuali Github repo. This could be an upgrade script, or the parameter names referenced in the QuestionResolver class could be changed to "Questionnaire ID" and "Question ID" to avoid needing any SQL changes at all.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/common/questionnaire/impl/QuestionResolver.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/common/questionnaire/impl/QuestionResolver.java
@@ -30,7 +30,8 @@ public class QuestionResolver implements TermResolver<Object> {
 
     public static final String MODULE_CODE = "moduleCode";
     public static final String MODULE_ITEM_KEY = "moduleItemKey";
-    public static final String QUESTIONNAIRE_REF_ID = "Questionnaire Ref ID";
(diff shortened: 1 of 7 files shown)


rojlarge:
Moved PR to contrib branch.

PR #1577, opened 2015-04-21
[KRACOEUS-7230] Re-implementing 5.2.1 QuestionResolver changes in KC 6.x

rojlarge:
KRACOEUS-7230

This re-implements some changes to the QuestionResolver code that we contributed before, and which made it into KC 5.2.1 but seem to have been left out of KC 6.0. It also includes a couple of our additional improvements/fixes to the QuestionResolver functionality, such as using the module sub item key when retrieving Answer Headers, filtering retrieved Answer Headers to only include those for the latest versions of Questionnaires, and making sure the Question Seq Id comparison is properly comparing String values. 

Please note that since the existing KRMS Question terms use "Question ID" and "Questionnaire ID" as parameters, the following SQL will need to be run to update them to the new "Seq ID" terminology:

```
update KRMS_TERM_RSLVR_PARM_SPEC_T set nm = 'Question Seq ID' where nm = 'Question ID';
update KRMS_TERM_RSLVR_PARM_SPEC_T set nm = 'Questionnaire Seq ID' where nm = 'Questionnaire ID';
update KRMS_TERM_PARM_T set nm = 'Question Seq ID' where nm = 'Question ID';
update KRMS_TERM_PARM_T set nm = 'Questionnaire Seq ID' where nm = 'Questionnaire ID';
``` 

Alternatively, the parameter names that QuestionResolver looks for could be changed to "Questionnaire ID" and "Question ID" to avoid the need for any SQL.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/coeus/common/questionnaire/impl/QuestionResolver.java
+++ b/coeus-impl/src/main/java/org/kuali/coeus/common/questionnaire/impl/QuestionResolver.java
@@ -30,7 +30,8 @@ public class QuestionResolver implements TermResolver<Object> {
 
     public static final String MODULE_CODE = "moduleCode";
     public static final String MODULE_ITEM_KEY = "moduleItemKey";
-    public static final String QUESTIONNAIRE_REF_ID = "Questionnaire Ref ID";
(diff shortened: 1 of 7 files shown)


leoherbie:
FYI. You can add new sql files if you want in the next logical sequence and I'll be sure to resequence them when I merge.

PR #1580, opened 2015-04-21
KRAFDBCK-10988: Adding totals to Negotiation Activity sub-tabs

rojlarge:
KRAFDBCK-10988

Adds item totals to the top of Negotiation Activity sub tabs.


rojlarge:
Whoops-- saw that this one was already merged into master.

PR #1584, opened 2015-04-28
[KRAFDBCK-12440] New parameter to hide Intellectual Property Review tab in IP

rojlarge:
KRAFDBCK-12440
This enhancement adds a new parameter "IP_INTELLECTUAL_PROPERTY_REVIEW_TAB_ENABLED" to determine whether or not the Intellectual Property Review tab should be shown in IP. If the parameter doesn't exist, then it defaults to showing the tab. 

Here is the SQL to add the new parameter: 

```
INSERT INTO KRCR_PARM_T (NMSPC_CD, CMPNT_CD, PARM_NM, OBJ_ID, VER_NBR, PARM_TYP_CD, VAL, PARM_DESC_TXT, EVAL_OPRTR_CD, APPL_ID)
VALUES ('KC-IP', 'Document', 'IP_INTELLECTUAL_PROPERTY_REVIEW_TAB_ENABLED', SYS_GUID(), 1, 'CONFG', 'Y', 'This param can be used to turn the Intellectual Property Review tab on/off in Institutional Propsoal', 'A', 'KC');
```

I'm still not sure what the best way to handle the SQL portion of our contributions, so please let me know what your preference is (aka adding a new folder to the coeus-db-sql module, or just including it in the PR so you can put it wherever you want).


leoherbie:
FYI.  You can add new sql files if you want in the next logical sequence and I'll be sure to resequence them when I merge.

PR #1585, opened 2015-05-19
[KRAFDBCK-10952] Adding ability to send Award Notice Notifications

rojlarge:
KRAFDBCK-10952
Added new notification type and logic for sending Award Notices from the print tab of the Award Actions page. This contribution also adds a new AWARD_NOTICE table to store persisted Notice PDFs, so that they can be easily retrieved from the notification recipients' Action List later.

I included the required SQL in both Oracle and MySQL dialects as V603_001__KRAFDBCK-10952.sql files in the respective coeus-db/.../data/migration/... folders, but as always, just let me know if there's another convention you'd like me to follow.


rojlarge:
Whoops, sorry-- this was supposed to be created against the contrib branch. I'll close and re-create.

PR #1586, opened 2015-05-19
[KRAFDBCK-10952] Adding ability to send Award Notice Notifications

rojlarge:
KRAFDBCK-10952
Added new notification type and logic for sending Award Notices from the print tab of the Award Actions page. This contribution also adds a new AWARD_NOTICE table to store persisted Notice PDFs, so that they can be easily retrieved from the notification recipients' Action List later.

I included the required SQL in both Oracle and MySQL dialects as V603_001__KRAFDBCK-10952.sql files in the respective coeus-db/.../data/migration/... folders, but as always, just let me know if there's another convention you'd like me to follow.


vikinzer:
This exists in the KualiCo system under RESSUP-86.  When this contribution is accepted that ticket should be pulled into the Kanban board for tracking on the appropriate release.

vikinzer:
This is now being tracked under RESKC-508 in the KualiCo queue.

leoherbie:
this have been accepted

leoherbie:
thanks so much for the contrib

PR #1588, opened 2015-07-10
Fix for "All My Proposals"

gouldner:
I worked with Brian Smith on this issue.

All my proposals should display proposals for which you are the initiator (you created the proposal).

This corrected version does that.  I removed the "Aggregator" lookup field because ultimately it was coded to actually lookup based on initiator so I renamed this lookup field to initiator and changed all the  corresponding code to match this name.

I also fixed a bug in the code that combined the lists of proposal numbers to search on after processing  proposalPerson, initiator, etc. 

This code now performs as it did in previous version of KC.

Please contact me if there are any questions.


leoherbie:
We have accepted this contribution and it will be in our next release.  Thanks so much.

gouldner:
FYI this merge conflicted when I pulled 1508.3.    Was this pull request Merged or did you just copy the changes by hand.   I know you needed to add Aggregator so I think you did the later.   But I think this is why I had merge conflicts.   If you merge the pull request and then make any modification you need it should have cleanly merged back to my version.

Just letting you know because this is the 2nd or 3rd time this has happened to me and Geo mentioned to me that it is happening to them as well.   So we may need to look into how the pull requests are being processed.

blackcathacker:
Out of curiosity how are you managing your local changes and new releases? I'm not sure how we can merge these PRs without you ending up with merge conflicts regardless, but maybe I'm not completely aware of everything git can do in this area. First, we can't actually merge these PRs directly due to the private repo we pull them into. We can cherry-pick commits, but I'm still not convinced that would help?

gouldner:
We forked the repo from kuali/kc
Our master is 100% unmodified and should always have no merge conflicts with upstream_kuali/master
We make our changes on uh_modifications.

Do deliver a change to you I work on a separate branch made from master.  This assures only the cherry picked change I make is merged and the rest is same as kuali/master

I push that up to my forked repo and generate pull request.

I understand what you are saying about how you are on private repo.   This is why I was against you not sharing access to that repo with us.   However knowing how git works it is possible to do this with no issues.   The key is for you to accept the pull request in kuali/kc (perhaps on a branch if you want  to vet it first) and then pull request it to your repo.   You can then merge to your repo and git will know it had the upstream source.   

When you feel the change is OK you can then merge the kuali/kc version to master again these extra steps only necessary because your KualiCo decision to make a private repo.  I realize this is extra work but it is a consequence of that decision.

Now when you push your changes out to kuali/kc git should recognize you are on the same baseline for this change and merge cleanly.

When I pull from the kuali_upstream source and merge kuali_upstream/master again git will recognize the source of this change already exists on my repo and merge cleanly.

Out of curiosity what approach are you using to accept these pull request.   If I knew the technique you are using I could probably point out what is being done wrong and causing the issues.  If you are hand copying the changes or creating a patch file and applying (same as hand merge) than I can promise this will always cause us pain.

PR #1590, opened 2015-10-08
MySql embedded mode fixes

chrisdenne:
Fixes issues with trying to build EMB mode schema using MySql scripts & a data conversion issue for EMB mode schema (both oracle & MySql).


leoherbie:
This is accepted.  I'm also fixing it for oracle.  Thanks!

PR #1593, opened 2015-12-29
[#RESSUP-504] New IRB and IACUC FYI process

rojlarge:
RESSUP-504

This contribution enables the Notify IRB and Notify IACUC actions to create new Fxxx documents off the base protocol like amendments and renewals. This allows FYIs to go through the full protocol lifecycle including committees and reviewers, and have their own dedicated questionnaires based on new sub module codes for IRB and IACUC. In this version of the enhancement, FYIs behave like amendments with only the Notes & Attachments module checked, but they don't merge anything back directly into the base protocol; instead, any attachments on the Notes & Attachments page of the FYI document are merged into the 'Notify IRB/IACUC' action in the history of base protocol as Action Attachments. This way a record of those attachments can be kept without affecting the actual documents intended to be associated with the protocol.

This new behavior can be turned on and off via the ALTERNATE_NOTIFY_IRB_ACTION_PARAM and ALTERNATE_NOTIFY_IACUC_ACTION_PARAM parameters.


rojlarge:
Note: V1508_001__FYI.sql and V1512_001__FYI.sql are identical and only one should have been committed-- feel free to ignore V1508_001__FYI.sql.

vikinzer:
Here is a link to a youtube video describing this functionality in some detail.  https://www.youtube.com/watch?v=76R9LCVi8Sc

vikinzer:
I believe this has been merged into the base code at least in as much as it's available in the demo environments.  I was curious when the github requests are closed relative to code inclusion.

PR #1596, opened 2016-01-06
[RESKC-1038] Change activity end date error to warning, and fix wording

gmcgrego:
RESKC-1038
Change from error to warning when Activity End Date is later than Negotiation End Date.  Change the wording of the warning message so it makes sense.  Also reference KRAFDBCK-10813.


leoherbie:
this PR has been accepted

PR #1597, opened 2016-01-19
[#RESSUP-1385] Added 'Document Status' field to Award lookup and results

rojlarge:
RESSUP-1385

This enhancement allows Awards to be looked up based on their VersionHistory status-- either "ACTIVE", "PENDING", or both. It also adds the display of this status to the Award lookup results, and also enables the Negotiation lookup to work properly with associated Awards in saved / PENDING version status.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/kra/award/dao/ojb/AwardLookupDaoOjb.java
+++ b/coeus-impl/src/main/java/org/kuali/kra/award/dao/ojb/AwardLookupDaoOjb.java
@@ -18,20 +18,28 @@
  */
 package org.kuali.kra.award.dao.ojb;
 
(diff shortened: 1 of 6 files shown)


vikinzer:
I put together some very draft documentation for this so that any schools that pull it down can see how it is meant to work.  It is located at https://docs.google.com/document/d/1-Ye3hkHrrwMxZrAUY0WhFCNsRhONjk4AYrE-Bf2vinI

rojlarge:
It looks like this was included in the 1604 release (associated with RESKC-1212, it looks like), so this can probably be closed, right?

blackcathacker:
Yes it was. Meant to do that.

PR #1598, opened 2016-01-20
[RESSUP-1391] Modifying Custom Attribute inquiry

brownp:
Modifying Custom Attribute inquiry to include information about the CustomAttribute and not the Custom Attribute Document


leoherbie:
This Pull Request has been accepted.

PR #1599, opened 2016-01-20
[RESSUP-1382] Adding Award Modification ID to Medusa

brownp:
Modifies "regular" medusa and Prop Dev's KRAD based Medusa to include the award's Modification ID.


tdurkin:
This has been accepted and will be part of the next release.

PR #1606, opened 2016-02-12
[#RESSUP-1455] Added Award lookup field for Institutional Proposal Number

rojlarge:
RESSUP-1455

This pull request adds an "Institutional Proposal ID" field to the Award lookup, which returns the active version of Awards associated with any version of the requested IP.

### DIFF
--- a/coeus-impl/src/main/java/org/kuali/kra/award/lookup/AwardLookupableHelperServiceImpl.java
+++ b/coeus-impl/src/main/java/org/kuali/kra/award/lookup/AwardLookupableHelperServiceImpl.java
@@ -24,10 +24,11 @@ import org.kuali.coeus.common.framework.person.KcPersonService;
 import org.kuali.coeus.common.framework.rolodex.Rolodex;
 import org.kuali.coeus.common.framework.unit.Unit;
 import org.kuali.kra.award.contacts.AwardPerson;
+import org.kuali.kra.award.dao.AwardLookupDao;
 import org.kuali.kra.award.document.AwardDocument;
 import org.kuali.kra.award.document.authorization.AwardDocumentAuthorizer;
 import org.kuali.kra.award.home.Award;
-import org.kuali.kra.award.dao.AwardLookupDao;
+import org.kuali.kra.award.home.fundingproposal.AwardFundingProposal;
 import org.kuali.kra.lookup.KraLookupableHelperServiceImpl;
 import org.kuali.rice.kew.api.KewApiConstants;
 import org.kuali.rice.kim.api.identity.Person;
@@ -42,6 +43,7 @@ import org.kuali.rice.krad.util.KRADConstants;
 import org.kuali.rice.krad.util.UrlFactory;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * This class provides Award lookup support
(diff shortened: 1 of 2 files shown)


coeus-impl/src/main/java/org/kuali/kra/award/lookup/AwardLookupableHelperServiceImpl.java
```diff
@@ -83,7 +86,17 @@
...
         setReferencesToRefresh(fieldValues.get(KRADConstants.REFERENCES_TO_REFRESH));
-        
+
+        if (StringUtils.isNotBlank(fieldValues.get(IP_NUMBER_FIELD))) {
+            List<String> awardNumbers =
+                    getBusinessObjectService().findMatching(AwardFundingProposal.class, Collections.singletonMap("proposal.proposalNumber", fieldValues.get(IP_NUMBER_FIELD)))
+                    .stream()
+                    .map(afp -> afp.getAward().getAwardNumber())
+                    .collect(Collectors.toList());
+            fieldValues.put("awardNumber", StringUtils.join(awardNumbers, '|'));
```
leoherbie: I'd probably use stream.reduce rather than stringutils.join

rojlarge:
Thanks for pointing that out Travis, that's definitely better than StringUtils.join. However, it looks like collecting with Collectors.joining is the cleanest option.

leoherbie:
I agree.  The concept is basically the same as a reduce in this case :)

coeus-impl/src/main/java/org/kuali/kra/award/lookup/AwardLookupableHelperServiceImpl.java
```diff
@@ -79,15 +83,28 @@
...
+        if (StringUtils.isNotBlank(fieldValues.get(IP_NUMBER_FIELD))) {
+            String awardNumbers =
+                    getBusinessObjectService().findMatching(AwardFundingProposal.class, Collections.singletonMap("proposal.proposalNumber", fieldValues.get(IP_NUMBER_FIELD)))
+                            .stream()
+                            .map(afp -> afp.getAward().getAwardNumber())
+                            .collect(Collectors.joining("|"));
+            if (StringUtils.isBlank(awardNumbers)) {
+                return new ArrayList<>();
+            }
+            fieldValues.put("awardNumber", awardNumbers);
```
leoherbie: If awardNumber is set as a search criteria then append.  Don't just overwrite the value with the map.put

blackcathacker: Related, @rojlarge did you try just renaming the field to fundingProposals.proposal.proposalNumber and not doing this search separately? That way hopefully OJB would do all the necessary joins to make it happen automatically?

rojlarge: Yeah, I was just looking at that... I swear there was a reason initially that we couldn't do that, but I just tried removing the changes to AwardLookupableHelperServiceImpl.java entirely (the lookup field is already named "fundingProposals.proposal.proposalNumber" in Award.xml) and it seems to work just fine. I can take another look at this tomorrow to make sure I'm not missing anything else, but doing it that way is way cleaner and handles the "awardNumber" case w/ wildcarding correctly.

rojlarge:
I did some more testing, and just using the correct field name works fine. I've updated this branch to only include the Award.xml change, and rebased it off of the latest master as well.

PR #1608, opened 2016-02-15
[RESSUP-1434] Prevent the AWARD.UPDATE_USER field from updating when versioned

brownp:
As seen on the Award History screen, when an award is versioned, the original version does not update the UPDATE_TIMESTAMP field, but does update the UPDATE_USER field. This pull request fixes that and makes it so neither the UPDATE_TIMESTAMP nor UPDATE_USER fields are updated on the old award version


leoherbie:
this PR has been accepted

